PR #20, opened 2016-11-17
move MultipleSchemaAnalyzer call into SchemaAnalyzer class

ismailsimsek:
adds analyzeMultipleSchemas function in SchemaAnalyzer and moves MultipleSchemaAnalyzer  logic into this class. with this change MultipleSchemaAnalyzer class is not needed. 

//if -all (evaluteAll) or -schemas given then analyzeMultipleSchemas  
//else analyze


rafalkasa:
Hi @ismailsimsek 

Your change is quite interesting me and I think that we can go in this direction to have analyzeMultipleSchemas as a method in class instead to have separate special class duplicating the code.

Your solution is working and generation for other database schemes are possible

Issue #46, opened 2017-03-09
WARNING exception in DbAnalyzerTest when schemaspy.properties is missing


rafalkasa:
Running org.schemaspy.DbAnalyzerTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.966 sec
Running org.schemaspy.ConfigTest
Mar 09, 2017 6:11:09 AM org.schemaspy.Config loadProperties
WARNING: schemaspy.properties (No such file or directory)
java.io.FileNotFoundException: schemaspy.properties (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at org.schemaspy.Config.loadProperties(Config.java:1622)
	at org.schemaspy.Config.fixupArgs(Config.java:1595)
	at org.schemaspy.Config.<init>(Config.java:141)
	at org.schemaspy.ConfigTest.testConfig(ConfigTest.java:16)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at junit.framework.TestCase.runTest(TestCase.java:176)
	at junit.framework.TestCase.runBare(TestCase.java:141)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:129)
	at junit.framework.TestSuite.runTest(TestSuite.java:252)
	at junit.framework.TestSuite.run(TestSuite.java:247)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:86)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:252)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:141)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:112)
[...]

PR #47, opened 2017-03-14
WARNING exception in DbAnalyzerTest when schemaspy.properties is missing #46

wkasa:
WARNING exception in DbAnalyzerTest when schemaspy.properties is missing #46


rafalkasa:
Looks Good I'm fine :-)

PR #57, opened 2017-04-03
Update teradata.properties

ClintEstwood:
Right now I have an error when I try to connect to the teradata DB:
Failed to connect to database URL [jdbc:teradata://myTeradataDB.NET/DBS_PORT=1025]

This URL should looks like that:
jdbc:teradata://myTeradataDB.NET/myTeradataDB,DBS_PORT=1025


rafalkasa:
@ClintEstwood thank you for your work and contribution. 
I'm not a Teradata DB specialist, so thank you for testing and using SchemeSpy with this solution.

PR #60, opened 2017-04-08
Svg images

wkasa:
Noticed, that font is not smooth when images was changed to svg.
Increasing font size +1 change this a little bit.


rafalkasa:
@wkasa now fonts look better when SchemaSpy is exporting them to the *.svg

PR #61, opened 2017-04-08
Update Maven to 3.5.0 (using Maven wrapper)

Riggs333:


rafalkasa:
@Riggs333  thank you for contributing and upgrade maven to the new version. Your help is always welcome :-)

PR #62, opened 2017-04-13
Support for Amazon Redshift database

wkasa:
Support for Amazon Redshift database.
Parameter for dbtype (-t) need to be set" -t redshift


rafalkasa:
@wkasa your help and contribution in to the project is invaluable. Thank you very much !

PR #63, opened 2017-04-13
Added imageformat parameter - png is default. svg as an option

wkasa:
After discusion with @rafalkasa we think, that png should be default image format and svg to be an option


rafalkasa:
I don't know why Graphviz generated svg diagrams with some issue in fonts. Sometimes table and colum names are generated incorrectly where diagrams format is set to svg. Thank you @wkasa for add this imageformat parameter option.

PR #68, opened 2017-04-23
Remove unused class ImageWriter, clean up ResourceWriter

Riggs333:
I couldn't find any usage of ImageWriter. And the method
ResourceWriter#writeResource was only used in ImageWriter so I deleted
it as well.
ImageWriter references the gif files that do not seem to be used
anywhere, so I also deleted them.


rafalkasa:
@Riggs333 thank you for your time and help make SchemaSpy better.

PR #74, opened 2017-04-27
Introduce Spring Boot, update maven plugins

Riggs333:
Switch from using 'plain' Spring Framework to **Spring Boot**.
Using Spring Boot's `CommandLineRunner` feature in `Main` class.
Therefore the `SchemaAnalyzer` can be autowired and the Spring context is
automatically created and initialized via the `@SpringBootApplication`
annotation.

With that change some **maven plugins** are updates to their latest version.

I also added a sample banner - as this is one of the coolest features in
Spring Boot. ;-)

**But:** I have to change the method `Config#isJarFile` to return `true` because
otherwise the resources cannot be found, like css and html template
files. I'm not sure what this check with the jar file is good for, it
didn't break anything, but we need to clarify or document what feature
it addresses.

Starting with Spring Boot there are several TODOs I see for the future:
We can get rid of the singletons and use Spring Beans / Components.
The `Config` class might be replaced by Spring's Configuration Properties.


src/main/java/org/schemaspy/Config.java
```diff
@@ -285,10 +285,12 @@ public String getTemplateDirectory() {
     }
 
     public boolean isJarFile() {
```
rafalkasa: This method was created by me to verify if application is execute as `java -jar schemaspy.jar`
I don't remember exactly the issue but it was something related to the layout resources.

Today I tested and everything is working as should :-) I will review the code and try to remove this method.

PR #75, opened 2017-04-29
schema and catalog comments are fetched and passed to template.

ismailsimsek:
This pull request adds two new (Schema, Catalog) classes. and pulls and passes schema and catalog comments to template.

two new parameters created to pull schema and catalog comments. used in pgsql.properties 
selectSchemasSql= used to get schema comment
selectCatalogsSql- used to get catalog comment 

additionally added schema and catalog sections to multi and main html pages which can be reverted or changed.

tested for Postgresql database.

Postgresql selects  
```
selectSchemasSql=SELECT nspname AS schema_name, pg_catalog.obj_description(oid, 'pg_namespace') AS schema_comment FROM pg_catalog.pg_namespace where nspname = :schema
selectCatalogsSql=SELECT datname AS catalog_name, pg_catalog.obj_description(datlastsysoid, 'pg_database') AS catalog_comment FROM pg_catalog.pg_database WHERE datname = :catalog ;

```


rafalkasa:
@ismailsimsek could you make screenshot to show example how will be look additionally added schema and catalog sections for your more complex database. 

Next thing is for which purposes this section should appear on main page. I would like to understand this change and the value that this change represents for other SchemaSpy users.

ismailsimsek:
Hi 
i will put sample screenshot soon.

ismailsimsek:
html change is done only in main.html and multi.html pages.
when there is comment for schema or catalog it will create red boxes in the screenshot.

and if schema or catalog doesn't have any comment or selectSchemasSql or selectCatalogsSql are not provided then this boxes not created. so for existing users there is no impact.

image

image

PR #80, opened 2017-05-04
Fix sonar issues

wkasa:
Fixex some sonar issues, mostly unnecessary boxing and unboxing


src/main/java/org/schemaspy/model/xml/TableColumnMeta.java
```diff
@@ -79,30 +79,35 @@
         digits = node == null ? 0 : Integer.parseInt(node.getNodeValue());
         
         node = attribs.getNamedItem("nullable");
-        isNullable = node == null ? false : evalBoolean(node.getNodeValue());
+        isNullable = node != null && evalBoolean(node.getNodeValue());
```
rafalkasa: @wkasa I'm not pretty sure if changed boolean expression will return the same results ? Could you verify or confirm this ? The best option will be if you can prepare some unit test.

PR #85, opened 2017-05-10
Adde spring-boot-starter-test

npetzall:
* Removed redudant dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,24 +64,6 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
-                <groupId>junit</groupId>
-                <artifactId>junit</artifactId>
-                <version>4.12</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
-                <groupId>org.assertj</groupId>
-                <artifactId>assertj-core</artifactId>
-                <version>3.6.2</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
-                <groupId>org.mockito</groupId>
-                <artifactId>mockito-core</artifactId>
-                <version>2.7.22</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
                 <groupId>com.github.spullara.mustache.java</groupId>
                 <artifactId>compiler</artifactId>
                 <version>0.9.1</version>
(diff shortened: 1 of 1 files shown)


Riggs333:
:+1: for using spring-boot-starter-test! :-)
I'd suggest to additionally [override the versions](https://spring.io/blog/2016/04/13/overriding-dependency-versions-with-spring-boot) of Mockito and AssertJ to the latest one.
The starter comes with some older versions.

I'm not sure, but I guess we also need to declare the dependencies that are used.
This is suggested (and marked as a warning) when you run `mvn dependency:analyze`:

```
...
[WARNING] Used undeclared dependencies found:
[WARNING]    org.springframework.boot:spring-boot:jar:1.5.3.RELEASE:compile
[WARNING]    org.mockito:mockito-core:jar:1.10.19:compile
[WARNING]    org.hamcrest:hamcrest-core:jar:1.3:compile
[WARNING]    org.assertj:assertj-core:jar:2.6.0:compile
[WARNING]    org.springframework:spring-beans:jar:4.3.8.RELEASE:compile
[WARNING]    junit:junit:jar:4.12:compile
[WARNING]    org.springframework:spring-core:jar:4.3.8.RELEASE:compile
[WARNING]    org.springframework:spring-context:jar:4.3.8.RELEASE:compile
[WARNING]    org.springframework.boot:spring-boot-autoconfigure:jar:1.5.3.RELEASE:compile
...
```

npetzall:
Would we even say that we want [versions-maven-plugin](http://www.mojohaus.org/versions-maven-plugin/)

Riggs333:
In what context do you want to use the versions plugin?

npetzall:
for the goals display-{dependency,plugin, property}-updates

npetzall:
Saw it late last night, intellij or I, hadn't updated the dependencies.

I'll override so it gets lastest minor of the version it had before the change.

Riggs333:
Thanks for updating the dependency versions.

Looking at the declared compile dependencies and reflecting on this issue I found this discussion:
https://github.com/spring-projects/spring-boot/issues/8341

So I guess the Spring Boot team has an argument here and I'm fine with it.
So relying on the spring boot starter would be ok for me.

rafalkasa:
@npetzall after the merge I don't see you as contributor to the SchemaSpy project could you check below explanation:

Sometime even if do fork, commit or merge your branch still it doesn't show your name in contributors. It may be due to the email which you set on command line is not mentioned in your git profile. Following steps work for me.

go to https://github.com/settings/emails
Add your email which is used on command line to 'add Email address' option
Now check the contributors tab on project. It will start showing your contributions.

https://help.github.com/articles/why-are-my-contributions-not-showing-up-on-my-profile/

npetzall:
@Riggs333 It's been a while since I've been doing maven. Mostly gradle and since dependencies are one-lines in gradle, you can declare them pretty compact.

There is a good thing about knowing whats really being used.
So having them declared is good, the version i still left to spring-boot.
But must say it takes up a lot of rows.

Riggs333:
@npetzall I completely agree with you.
I think we can leave it like that for now. If later the `pom.xml` becomes bigger or more complex we can address it then and see what suits us.

Issue #71, opened 2017-04-24
MariaDB 5.5.x support? 


GETandSELECT:
I use MariaDB (not Oracle MySQL) and can't connect. Password works with `mysql` client (in `.my.cnf `hardcoded) and has no special characters, only alphanum.

I use RHEL7 VM.

```
# /usr/java/jre1.8.0_77/bin/java -jar /root/tmp/schemaspy-6.0.0-rc1.jar
Apr 21, 2017 3:56:32 PM org.springframework.context.annotation.AnnotationConfigApplicationContext prepareRefresh
INFO: Refreshing org.springframework.context.annotation.AnnotationConfigApplicationContext@30dae81: startup date [Fri Apr 21 15:56:32 CEST 2017]; root of context hierarchy
Using database properties:
  [/root/tmp/schemaspy-6.0.0-rc1.jar]/org/schemaspy/types/mysql.properties

Failed to connect to database URL [jdbc:mysql://localhost:3306/sbmongodb]

java.sql.SQLException: Invalid authorization specification message from server: "Access denied for user 'root'@'127.0.0.1' (using password: YES)"
        at com.mysql.jdbc.MysqlIO.checkErrorPacket(MysqlIO.java:2001)
        at com.mysql.jdbc.MysqlIO.checkErrorPacket(MysqlIO.java:1907)
        at com.mysql.jdbc.MysqlIO.secureAuth411(MysqlIO.java:2524)
        at com.mysql.jdbc.MysqlIO.doHandshake(MysqlIO.java:818)
        at com.mysql.jdbc.Connection.createNewIO(Connection.java:1808)
        at com.mysql.jdbc.Connection.<init>(Connection.java:452)
        at com.mysql.jdbc.NonRegisteringDriver.connect(NonRegisteringDriver.java:411)
        at org.schemaspy.DbDriverLoader.getConnection(DbDriverLoader.java:66)
        at org.schemaspy.service.SqlService.connect(SqlService.java:65)
        at org.schemaspy.SchemaAnalyzer.analyze(SchemaAnalyzer.java:197)
        at org.schemaspy.SchemaAnalyzer.analyze(SchemaAnalyzer.java:86)
        at org.schemaspy.Main.main(Main.java:52)
15:56:33.543 WARNING: Main.main - Connection Failure
org.schemaspy.model.ConnectionFailure: java.sql.SQLException: Invalid authorization specification message from server: "Access denied for user 'root'@'127.0.0.1' (using password: YES)"
[...]

PR #86, opened 2017-05-10
Include nar when loading dp sibling libraries

npetzall:
loadjars will include .?ar (nar,war,jar), H2MemoryRule for jUnit
* H2Memory rule with script support I believe
* H2 dependency with scope test
* Config constructor change from String[] argv to String... argv

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,11 @@
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
         </dependency>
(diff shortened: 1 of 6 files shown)


npetzall:
relates to issue #71 
nar is a native library packaging

rafalkasa:
Could you resolve the conflicts in the pom.xml before I will made the merge with master branch

npetzall:
Will do tonight

PR #87, opened 2017-05-10
Refactoring Config: Introduce JCommander to parse CLI arguments

Riggs333:


npetzall:
It would be a breaking change but Spring boot already supports command-line arguments.
https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-command-line-args

aadrian:
+1 for JCommander - much better than Spring boot.

Riggs333:
@npetzall Yes, I know this feature and like it very much.
But as you mentioned this would be a breaking change to rely only one that Spring Boot handling of arguments/properties.

With JCommander we can easily define aliases for command line arguments and also we are not bound to use double dash arguments.
Also JCommander offers a nice [usage print message](http://jcommander.org/#_usage). And I like the error messages of missing required arguments more than.

npetzall:
Looks a bit like args4j, will the CommandLineArgs replace Config?

aadrian:
@npetzall much better and easier to use than args4j :)

Riggs333:
To be honest I didn't really evaluate a lot of CLI parser libraries. JCommander was one of the first I looked into and found sufficient and acceptable. I took a short look at commons-cli but find it inferior to JCommander.

Yes, my plan was to replace the `Config` class or at least all its handling of command line arguments with the new class `CommandLineArgs`. (Although I'm not yet sure about the name.)
There is some logic in the `Config` class that I see to be moved to other classes, but I guess it will be incremental steps to tame this beast. ;-)

npetzall:
I'm not proposing args4j, only that it binds arguments to a class which lead me to the question with the Config replacement.

Config and logic doesn't belong in the same class so moving them sounds great from my point of view.

Riggs333:
I did some changes to the PR. Now this all definitely needs some review. :-)
Before digging into this config stuff alone for too long and loosing track I'd like you to take a look.

My main issue was to **replace the existing `Config` class** (at least in the long term) with something that is more readable and scalable, and has fewer lines of code.
The choice fell on [JCommander](jcommander.org).

There are several points I was or still am struggling with:
- parsed command line arguments shall be available in the **Spring context** (as a bean)
- handling of the `configFile` parameter to **externalize configuration properties** (with fallback to an optional `schemaspy.properties` file)
- providing **readable information** to the user for usage of arguments and in case of errors

I'd be happy and grateful to get your feedback on this. :-)

npetzall:
It's been a while since I used Spring so I actually don't remeber execution orders and such.
I say shouldn't something like this be possible
```
@Configuration
public class SchemaSpySpringConfiguration {

  @Autowired
  private ApplicationArguments applicationArguments;

  @Bean
  public CommandLineArgs create(applicationArguments) {
     //create and parse CommandLineArgs using applicationArguments.getSourceArgs()
    return commandLineArgs
  }
}
```

Riggs333:
@npetzall Great! This seems to work.
Will do some more tests locally and then update the PR. Thanks!

Riggs333:
I updated the files, introduced a `SchemaSpyConfiguration` class, did some refactoring.

npetzall:
Like how this fixes #91

Riggs333:
@npetzall yes, at least this might be a start.

@rafalkasa please take a look at the changes and decide if it is already enough to merge. Or let me know what needs to be changed.

After the merge I can think of small PRs following for every config parameter that will be moved from the `Config` class to `CommandLineArguments` (or something similar depending on the usage).

rafalkasa:
@Riggs333 before I will check changes made in code I verified how SchemaSpy is working after the changes.
Unfortunately on Windows all the paths are retrieved incorrectly eg. c:\toolsschemaspypostgresql-42.0.0.jar
Could you verify this before I will merge your PR ?

rafalkasa:
Small update paths are retrieved incorrectly only for Windows format paths eg. `c:\tools\schemaspy\postgresql-42.0.0.jar` when the param will be change to the unix paths `c:/tools/schemaspy/postgresql-42.0.0.jar` everything working correctly also on Windows environment.

I would like to have backward compatibility otherwise we will generate a lot of issues and we will disappoint the Windows users.

Riggs333:
@rafalkasa Thank you for testing this.
I will take a look today and try to fix it.

Riggs333:
I fixed some file handling in `MustacheWriter` and replacing backslashes in `ConnectionURLBuilder` because that didn't work out for me in Windows.
Now the following command works for me (in Windows 7 Enterprise):

```bash
java -jar schemaspy-6.0.0-rc1.jar -t h2 -dp C:\schemaspy\h2-1.4.194.jar -db C:\schemaspy\mydb -u sa -s PUBLIC -o report
```

rafalkasa:
@Riggs333 I try to test your change but code doesn't change. Did you push changes to github branch ?
Maybe I'm doing something wrong but from my perspective everything working as before...

Riggs333:
@rafalkasa Yes, I pushed the changes. This is strange.
Can you provide here the complete command line call like `java -jar schemaspy.jar -dp ...`?

I wonder why the driver path (`-dp ...`) is not recognized correctly on your machine.
In `CommandLineArguments` the `driverPath` value is not yet used anywhere.
I'll try to figure out what has changed.
Does your scenario run with the master branch?

rafalkasa:
@Riggs333 i know where the issue is. I'm using `schemaspy.properties` file instead application command line parameters with configuration as below:

```
# type of database. Run with -dbhelp for details
schemaspy.t=pgsql

# optional path to alternative jdbc drivers. 
schemaspy.dp=c:\tools\schemaspy\postgresql-42.0.0.jar


# database properties: host, port number, name user, password
schemaspy.host=localhost
schemaspy.port=5432
schemaspy.db=library
schemaspy.u=postgres
schemaspy.p=

# output dir to save generated files
schemaspy.o=c:\tools\schemaspy\pgsql\library

# db scheme for which generate diagrams
schemaspy.s=public

# Uncoment to generate diagrams for all schemas in database
#schemaspy.all

# Uncoment to skip creating diagrams for views
#schemaspy.noviews
#schemaspy.loadjars=true
schemaspy.template=c:\develop\layout\
```
I would confirm that everything working correctly if you use application command line parameters 😄 

You can read more about this configuration file here: http://schemaspy.readthedocs.io/en/latest/started.html

Riggs333:
@rafalkasa thanks for the verification and explanation.
I confirm the was a bug: When reading the properties from the properties file the backslashes were lost.
This should be fixed now. I made changes to `PropertyFileDefaultProvider` where the handling of windows paths is restored.

rafalkasa:
@Riggs333 I would like confirm that you fixed the issue with windows path in `schemaspy.properties` file, but I found out next few new issues 😄  

1. When I run SchemaSpy form IntelliJ IDEA `schemaspy.properties` doesn't exist in the main project folder and parameter provided as below:

    `dp c:\tools\schemaspy\postgresql-42.0.0.jar -o c:\tools\schemaspy\pgsql\library -t pgsql -host localhost -port 5432 -db library -u postgres -p test -s public`

    such error appear:
    `Writing/graphing summary.......2017-05-18 08:12:12.865  WARN 364 --- [           main] org.schemaspy.Main                       : Bad config Unable to find requested file: relationships.html in directory /layout`

2. The same case as above but `schemaspy.properties` exist in the main project folder everything working correctly, but is something like a mix part configuration is read from command line arguments other part from `schemaspy.properties` file eg. password was retrieved from the file. In my opinion we should read the configuration only from one source it can be `schemaspy.properties` or command line arguments. Mixed version can be strange and confused.

3. I don't provide any command line parameters only `schemaspy.proprties` file in main folder. SchemaSpy recognize that file exist, but instead take the configuration from the file help information is presented.

4. Below configuration run from cmd is working:

    `java -jar schemaspy-6.0.0-rc1.jar -dp c:\tools\schemaspy\postgresql-42.0.0.jar -o c:\tools\schemaspy\pgsql\library -t pgsql -host localhost -port 5432 -db library -u postgres -p test -s public`

5. Below configuration is working too:

    `java -jar schemaspy-6.0.0-rc1.jar -configFile schemaspy.properties` this configuration also work from  IntelliJ IDEA

Users of SchemaSpy can run application in few different cases and both of them should work as before. I don't know how to test this scenarios by integration tests but I know that @npetzall start to doing something. Maybe all together we will find the way how build the tests 😄 

@Riggs333 thank you for your work and contribution I really appreciate that !

npetzall:
@rafalkasa should schemaspy.properties be  autodiscover on classpath?
1, Might this be a classpath issue in intellij?
2, mixed should be supported. Configfile first and overriden by cmdArgs. Is my opinion.

I'll see what can be done in tests.

wkasa:
http://schemaspy.readthedocs.io/en/latest/started.html#configuration
It was intended from the beginning that command-line parameters will override those configured in schemaspy.properties file.

rafalkasa:
@wkasa but I observed that is opposite schemaspy.properties override the command-line parameters e.g. issue with password

rafalkasa:
@npetzall ad1 - In my opinion this is something according to the changes in the PR before everything working as should no such issue exist.

Riggs333:
Thanks! I'll check the issues this evening.

Riggs333:
I pushed some changes to address the issues.

1. Running SchemaSpy from inside IntelliJ IDEA should be working
The configuration lookup for command line arguments and/or config files works the same way as via the command line call (`java -jar ...`)

2. The lookup for configurations should be working as described in the documentation (as @wkasa mentioned). If the user provides  `-configFile=my.properties` this file will be picked up. If the user additionally provides command line arguments then these will have priority over those in the file. Nevertheless they can be combined. Anyway: required properties that are neither arguments and not in the file will exit the app. If there exists a file `schemaspy.properties` in the same folder or in the user's home directory this will be used for the lookup as a fallback.

3. I added some logging to inform the user that a configFile is found and used.

@rafalkasa I try to reproduce the issue that a password from the configuration file has priority over the command line argument.

Riggs333:
Since the **handling of the password** parameter is a little more complex (like prompting for password, fallback to reading from system environment variable `schemaspy.pw`, etc. see `Config` class) I do **not** let JCommander parse the password parameter, yet. (See the commented out code in `CommandLineArguments` class.)

So the behaviour should now fall back to the original one.

rafalkasa:
@Riggs333 I have done verification of your last changes you will find my feedback below:

1. Feature that you can recognize or not `schemaspy.properties` doesn't work when I run SchemaSpy from command line call like `java -jar schemaspy-6.0.0-rc1.jar` or from IntelliJ IDEA you can look below at example output of course schemaspy.properties exist in the same folder:

    ```
    [34mINFO [0;39m - Starting Main v6.0.0-rc1 on DESKTOP-JVOBE0G with PID 12108 
    (C:\develop\schemaspy\target\schemaspy-6.0.0-rc1.jar started by rafal in 
    c:\develop\schemaspy\target)
    [34mINFO [0;39m - No active profile set, falling back to default profiles: default
    [1;31mERROR[0;39m - The following options are required: [-u | --user | user | schemaspy.u | 
    schemaspy.user], [-o | --outputDirectory | outputDirectory | schemaspy.o | schemaspy.outputDirectory]
   ```

    Everything is working when `-configFile` parameter is provided both from cmd and IDEA eg. when application doesn't find provided file `ERROR - Could not find config file: schemaspy.properties` appear. 

    Unfortunately SchemaSpy can't recognize that file exist or not in the same folder and take the config parameters automatically from the file, instead of is informing that some of required parameters was not provided.

2. Overriding by command line arguments parameters exist in the file also is working correctly
3. Could you check because after when you extend the logging information quite strange character star appearing on Windows cmd each time when you log start new line ``

4. Password parameter is working the same as original one, what is also very good

Riggs333:
@rafalkasa Thanks for your feedback. I will take a look into the issues tonight.

Riggs333:
@rafalkasa I fixed the issues you mentioned:
1. `schemaspy.properties` is now picked up if it is found (if `-configFile` parameter is left out)
3. Logging under Windows should no more print those strange characters.

Riggs333:
Thank you very much, @rafalkasa!
I'll try to keep the next PRs smaller. :-)

Issue #26, opened 2017-01-03
Unable to find layout/schemaSpy.css 


saihegde:
Writing or graphing the summary out fails with the error message 

Unable to find requested file schemaspy-6.0.0-beta.2.jar!\layout\schemaSpy.css

rafalkasa:
Hi @saihegde I confirm this bug it will be fixed today

rafalkasa:
issue fixed

tjdavis3:
This error currently exists again in 6.0.0-rc1:

$ java -jar schemaspy-6.0.0-rc1.jar
May 11, 2017 12:20:19 PM org.springframework.context.annotation.AnnotationConfigApplicationContext prepareRefresh
INFO: Refreshing org.springframework.context.annotation.AnnotationConfigApplicationContext@4edde6e5: startup date [Thu May 11 12:20:19 MDT 2017]; root of context hierarchy
Using database properties:
  [schemaspy-6.0.0-rc1.jar]/org/schemaspy/types/pgsql.properties
Gathering schema details................................................(192sec)
Connecting relationships................................................(2sec)
Writing/graphing summary..12:23:35.007 WARNING: Main.main - Bad config Unable to find requested file: file:/Users/.../schemaspy-6.0.0-rc1.jar!/layout/schemaSpy.css

Riggs333:
This issues has the same root cause as #83

I'll try to provide a fix tommorrow as long as noone else does until then.

rafalkasa:
@Riggs333 now I remember why this isJarFile method was added to the Config class.

You can understand more when you will analyzing my commit  71e32e7 which is fixing the problem reported once again by @tjdavis3.

rafalkasa:
@tjdavis3 I try to reproduce the issue and after few tests I can't.
I verify schemaspy lunched it from InteliJ IDEA, from command shell and still everything is working correctly.

Your schemaspy-6.0.0-rc1.jar is build by you on your local machine or maybe downloaded version from [schemaspy 6.0.0-rc1](https://github.com/schemaspy/schemaspy/releases/download/v6.0.0-rc1/schemaspy-6.0.0-rc1.jar)

Third version can be [schemaspy@latest](https://github.com/schemaspy/schemaspy/raw/gh-pages/schemaspy-6.0.0-rc1.jar)

Could you tell as which version did you use ?

rafalkasa:
[...]

PR #90, opened 2017-05-12
Fix use of -css command line argument. Remove unused method Config#isJarFile

Riggs333:
issue #26

I'm still not quite sure if the behaviour is correct as intended.

Here's what I observerd so far regarding the priority of used css file:
Given command line argument `-css my.css` (default: `schemaSpy.css`):

1.  look for css file in same folder as jar
2.  look for css file in template directory (default '`layout`')
3.  look for css file in user.dir (home dir)
4.  use internal css file in classpath/schemaspy jar file (`src/main/resources/layout`)
 
@rafalkasa please take a look if this is how it is supposed to be.

### DIFF
--- a/src/main/java/org/schemaspy/Config.java
+++ b/src/main/java/org/schemaspy/Config.java
@@ -268,15 +268,6 @@ public class Config {
         return templateDirectory;
     }
 
-    public boolean isJarFile() {
-        return true;
-        //FIXME what is this good for?
-//        if (jarFile == null) {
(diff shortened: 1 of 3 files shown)


rafalkasa:
@Riggs333 because problem `Bad config Unable to find requested file: \layout\schemaSpy.css` still occur in this PR could you first merge it with master. I would like to avoid any problems and merge something that is not working correctly. Thank you in advance.

PR #92, opened 2017-05-14
Simple integration test of DatabaseService

npetzall:
* Added maven-failsafe-plugin
* Added dbType for h2memory in test resources
* Added getter to H2MemoryRule to access the keepAlive connection for support in test


npetzall:
So verify goal isn't run in the Travis-CI, only package. So failsafe plugin isn't triggered and integration tests are not run. One thing would be to change to verify. Second I wonder if jacoco works out of the box.

npetzall:
Last commit addresses issue #93

PR #96, opened 2017-05-31
Docker image builds: snapshot and relase

npetzall:
# Common
They current don't contain any jdbc-drivers.

# Release
* Release contains a small entry script since we have changed the main-class since rc1.
* Release will download the github release artifact.

# Snapshot
* Will git clone the repository and build it.

### DIFF
--- a/docker/release/Dockerfile
+++ b/docker/release/Dockerfile
@@ -1,0 +1,37 @@
+FROM openjdk:8-jre-alpine
+
+ENV LC_ALL=C
+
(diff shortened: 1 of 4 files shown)


npetzall:
Do we want to add for instance mysql, mariadb, postgresql, jTDS as default drivers I believe we will cover a majority?

rafalkasa:
I believe that it will be good add the most used jdbc drivers mention above

PR #104, opened 2017-06-05
Docker images v2

npetzall:
New docker file that is always used for all docker image builds.
Contains mysql, mariadb, jdts and postgreql driver by default.
Expected to be built by travis-ci and pushed to docker registry.

I will skip building until secure envs are added to travis-ci: DOCKER_USER and DOCKER_PASS

We need to remove the automatic build and replace with a normal repo. The documentation for the images will be hosted in that repo.

tags:
snapshot -> Built from master
release -> Built from master with a v[n].[n].[n].* tag
latest -> will point towards latest release including RC


rafalkasa:
@npetzall changes merged with master branch and secure envs have been added to travis-ci.

PR #131, opened 2017-08-07
Issue 109 dbtypes

npetzall:
Added a constructed value for option `<hostOptionalPort>`

So port is always option.
Separator between host and port can be customized.

Should solve 109.

I.e `-host localhost:3306` and `-host localhost -port 3306` creates the same result.


rafalkasa:
@npetzall the change looks awesome, thank you for your contribution and whole work that you did.
I would like test and review this change carefully before merge. Grate that you prepare unit test, excellent work.

Issue #144, opened 2017-09-01
sortable columns in modules TABLES, COLUMNS and CONSTRAINTS


mapausch:
In 6.0.0-RC1 the tables were sortable (ascending/descending) with a click on the column headers.
In schemaspy@latest this doesn't work anymore.
Also the buttons "All", "Tables", "Views" and "Comments" disappiered somewhere between RC1 and schemaspy@latest:

from 6.0.0-RC1:
![grafik](https://user-images.githubusercontent.com/18686836/29966910-47ebde20-8f14-11e7-882a-cb4e615f8dcb.png)

from schemaspy@latest:
![grafik](https://user-images.githubusercontent.com/18686836/29966990-aa0adb92-8f14-11e7-8209-02536f6031bf.png)

rafalkasa:
@mapausch 

Unfortunately I can't reproduce the same bug that you have. 

Could you open in the web browser Developer tools and look to the console logs ? 
Maybe there is something which help us.

Thank you for using SchemaSpy and create the issue.

mapausch:
I think I found the reason, why it doesn't work for me but works for you:

We generate our SchemaSpys on a Linux server and share the directory via a webserver, so our users don't need to use SchemaSpy on their local (Windows) machines.
When I copy the directory from the Linux server to my Windows machine than I can sort the columns.

In the console logs of the browser I can see that a HTTP GET request for "jQuery-2.2.3.min.js" (with a uppercase Q) gets replied with a 404 (File not found) by the webserver.
The 404 reply is generated because in the path on the webserver the file is named "jquery-2.2.3.min.js" (with a lowercase q). 

So I assume that it doesn't work from the webserver, because the underlaying filesystem in Linux is case-sensitive. And it works from Windows because there the filenames are case-insensitive.

In the screenshot you can see in the upper part the file name on the linux server and in the lower part the output of the webbrowser console:
![grafik](https://user-images.githubusercontent.com/18686836/30688795-6b8a719c-9ebf-11e7-9167-ab3559e2c1a6.png)

mapausch:
https://github.com/schemaspy/schemaspy/pull/146

rafalkasa:
Thank you for your time and investigation the issue. Your contribution to the SchemaSpy is very appreciated.

rafalkasa:
@mapausch as I understand everything is working now for you issue can be closed

PR #146, opened 2017-09-21
Rename jquery-2.2.3.min.js to jQuery-2.2.3.min.js

mapausch:
https://github.com/schemaspy/schemaspy/issues/144


rafalkasa:
@mapausch thank you for contributing to the project. 

Could you change in your PR src/main/resources/layout/container.html line where jQuery is added to the webpage instead changing name of the files.

All web dependencies are managed by bower and I am afraid that name of the file can change without our control eg. when we upgrade admin-lte to new version.

mapausch:
@rafalkasa you're right, changing the filename was the worst of 2 options :-)

changed the PR, but didn't find the Revert Button to revert the commit that changes the filename so I did another commit that just renames the file back to the old name... maybe some day I learn how to use git/GitHub ;-)

Issue #136, opened 2017-08-17
Page details in HTML title tag


Mihailoff:
Right now all generated pages have same title 
`<title>XYZ Database</title>`

I wonder if there a way to have more contextual information out there, for example, on ABC table page have a title like `<title>ABC table, XYZ Database</title>`.

Also, it would be good to have an ability to add default part of title so on every page title would be `<title>XYZ Database / Text for every page</title>`

PR #148, opened 2017-09-25
Add table name in to HTML title tag

rafalkasa:
This PR add contextual information to the tables pages for example ABC table page will have a title like <title>ABC table - XYZ Database</title>.

This PR fix issue #136


npetzall:
Checked it, haven't worked with mustache but now I understand the block thing so looks good to me too.

rafalkasa:
@npetzall, @Riggs333 thank you for review I will merge the code

PR #150, opened 2017-09-29
Update dependencies Spring Boot, AssertJ and Mockito to latest versions

Riggs333:


npetzall:
Haven't checked  but awesome, this is one thing that's missing at every place I worked.

PR #151, opened 2017-09-29
Display on main page database description in HTML

### LINKED ISSUE #149: command line parameter "-desc"
in the "old" SchemaSpy from sourceforge I could add a command line parameter "-desc" to add additional information to the generated SchemaSpy HTML files.
I used it to add information about hostname (or IP), username and schemaname to the HTMLs:

grafik

To get this I inserted some HTML to the description like that:
`java -jar schemaspy.jar ....  -desc "<b>Connection details:</b><br>Host/Port: <b> ... <br>DBName: <b>..."`

We manage a lot of databases and schemas for production, test and QA for different customers and internal projects and for our users sometimes it's not easy to recognize from which database the generated files are. This additional information did help us a lot, so would be great if there would be a possibility to have something like "-desc" again.

Currently when I add "-desc" than there's no error or warning about a "unknown parameter" but the description also does not appear in the HTML.

rafalkasa:
This pull request is covering backward compatibility with SchemaSpy 5.0.x after information provided after `-desc` parameter will be visible on main page fix #149

`-desc "<b>Connection details:</b></br><b>Host/Port: </b>127.0.0.1:1433</br><b>DBName: </b>Librrary</br>"`

schemaspy-description

### DIFF
--- a/src/main/java/org/schemaspy/view/HtmlMainIndexPage.java
+++ b/src/main/java/org/schemaspy/view/HtmlMainIndexPage.java
@@ -24,7 +24,6 @@ import java.util.*;
 
 import org.schemaspy.DbAnalyzer;
 import org.schemaspy.model.Database;
-import org.schemaspy.model.ForeignKeyConstraint;
 import org.schemaspy.model.Table;
 import org.schemaspy.util.Markdown;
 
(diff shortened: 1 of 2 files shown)


rafalkasa:
Sorry that I will not follow the rules but I would like to include this change in today release

PR #166, opened 2017-11-01
Fix some sonar warnings, consistent use of try-with-resource for SQL

danwatt:
Fix the following Sonar issues to try and get the quality up a little bit (and learn the codebase in the process)

* Removed a dead store (1 occurrence)
* Use isEmpty() to check whether the collection is empty or not. (4 occurrences)
* ...  returns a string, there's no need to call "toString()". (6 occurrences)
* Use try-with-resources for Statements and ResultSets. There were a few different patterns to how these were being closed - sometimes null checked without try-catch around the close, sometimes in the wrong order (statement then resultset), sometimes they were not null checked, and sometimes with a stacktrace being printed.
* There were a couple of `while ... break` loops that executed at most one time. Changed these to `if` blocks. This was an Intellij warning.

I debated implementing a closeQuietly method (Spring provides one, but I didn't want to introduce a dependency on Spring that deep into the project), and erred on the side of using built-in functionality instead.

### DIFF
--- a/src/main/java/org/schemaspy/service/DatabaseService.java
+++ b/src/main/java/org/schemaspy/service/DatabaseService.java
@@ -84,42 +84,31 @@ public class DatabaseService {
    private void initCatalogs(Config config, Database db, ProgressListener listener) throws SQLException {
 
             String sql = Config.getInstance().getDbProperties().getProperty("selectCatalogsSql");
-            PreparedStatement stmt = null;
-			ResultSet rs =  null;
+
             if (sql != null && db.getCatalog() != null) {
-                try {
-                stmt = sqlService.prepareStatement(sql,db,null);
-                rs = stmt.executeQuery();
-                    while (rs.next()) {
+                try (PreparedStatement stmt = sqlService.prepareStatement(sql, db, null);
+                     ResultSet rs = stmt.executeQuery();) {
+                    if (rs.next()) {
                          db.getCatalog().setComment(rs.getString("catalog_comment"));
-                         break;
                     }
                 } catch (SQLException sqlException) {
                     //db.getSchema().setComment(null);
-                } finally {// 
-                    stmt.close();
-                  	rs.close();
                 }
             }
     }
 
     private void initSchemas(Config config, Database db, ProgressListener listener) throws SQLException {
     	  String sql = Config.getInstance().getDbProperties().getProperty("selectSchemasSql");
-          PreparedStatement stmt = null;
-			ResultSet rs =  null;
+
           if (sql != null &&  db.getSchema() != null) {
-              try {
-              stmt = sqlService.prepareStatement(sql,db,null);
-              rs = stmt.executeQuery();
-                  while (rs.next()) {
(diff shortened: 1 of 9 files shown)


danwatt:
Working on fixes... just got my machine running the docker-dependent tests and I was able to reproduce the error

npetzall:
Ohh, yeah, was it troublesome to get the docker tests running? Do we need instructions for that?

npetzall:
I've created the docker test thing and don't assume they are a 100% correct since they are created after and not before implementation.

danwatt:
I will submit a PR in the future, maybe modifying the `TESTING.md` file with some instructions. I hastily got docker running on this computer, Ill do a more thorough job on another computer and keep track of what I had to do

danwatt:
Ok, fixed the issue - there was one place that a try-with-resources actually was not needed, because it needed to return an open statement.

src/main/java/org/schemaspy/service/DatabaseService.java
```diff
@@ -84,42 +84,31 @@ public void gatheringSchemaDetails(Config config, Database db, ProgressListener
...
-                rs = stmt.executeQuery();
-                    while (rs.next()) {
+                try (PreparedStatement stmt = sqlService.prepareStatement(sql, db, null);
+                     ResultSet rs = stmt.executeQuery();) {
+                    if (rs.next()) {
                          db.getCatalog().setComment(rs.getString("catalog_comment"));
-                         break;
                     }
                 } catch (SQLException sqlException) {
                     //db.getSchema().setComment(null);
```
rafalkasa: @danwatt instead leave this `catch `empty as it was could you change the code to something like:

```
String msg = listener.recoverableExceptionEncountered("Failed to retrieve select catalog sql query", sqlException, sql);
if (msg != null) {
     logger.warning(msg);
}
```

another option maybe better, it can be add something like:

```
logger.log(Level.SEVERE, selectCatalogsSql);
throw sqlException;
```

src/main/java/org/schemaspy/service/DatabaseService.java
```diff
@@ -84,42 +84,31 @@ public void gatheringSchemaDetails(Config config, Database db, ProgressListener
...
-                  while (rs.next()) {
+              try (PreparedStatement stmt = sqlService.prepareStatement(sql, db, null);
+                   ResultSet rs = stmt.executeQuery()) {
+
+                  if (rs.next()) {
                        db.getSchema().setComment(rs.getString("schema_comment"));
-                       break;
                   }
               } catch (SQLException sqlException) {
                   //db.getSchema().setComment(null);
```
rafalkasa: The same situation is here, I know that it was before but it will be good to not leave in the code empty `catch`

danwatt:
@rafalkasa updated
Going with exceptions - if the schemas or catalogs cannot be initialized properly, that seems to be worth throwing an exception.

rafalkasa:
@danwatt Yes this is true, I'm afraid that if we will not have this logged and not throw the exception. The code will work not deterministic, and we will have error but it will never appear. If this change start creating unexpected issues we have to redesign this part of the code.

Issue #162, opened 2017-10-26
show columns page by page and not all at once


mapausch:
in the tables overview there's a control to scroll tables page by page:
![grafik](https://user-images.githubusercontent.com/18686836/32052917-e731f788-ba59-11e7-9e3c-ddd4406755d6.png)

We have database schemas that have about 12k columns across about 850 tables and views.
The columns.html hast just about 3mb in size but rendering of all columns in the schema in the browser at once takes some time.

So a control like that in the tables overview maybe could prevent the browser to render everything at once and make a click on the COLUMNS module more responsive.

npetzall:
@mapausch @rafalkasa @danwatt Is this finished? Everybody happy?

npetzall:
Closing, if this is not ok. Feel free to re-open.

mapausch:
The initial processing of columns.html still is very slow.
Browsing the pages is fast.

If we could have both would be perfect :-)

But don't know if it's even possible to implement that on initial load only the first "page" is rendered :-/

npetzall:
We could have a look att using a browser database or similar. We could fake it by only adding first page in html and when rendering completed lazy load rest of the content. Only important thing would be to have the same sorting in html and Java.

npetzall:
I'm gonna re-open this.
http://madebyknight.com/optimizing-datatables-performance/

Load json instead of dom and possibly the deferred render could make the load time much better.

npetzall:
@mapausch I've tested adding a json object instead of directly in the dom, I turned on deferedRender and I went from 17.83 seconds down to 1.79 for 6661 columns/rows.

It's late right now and I know that things can be done. So I'm gonna polish the solution in the next days and also see if the current implementation isn't just too ugly.

PR #168, opened 2017-11-01
#162 - Add pagination

danwatt:
#162 - Add pagination to both the column and constraint pages

Follow the setting used on the main table listing page

Tested on a database with ~800 tables, ~9,000 columns, and ~600 constraints.

I do wonder if adding pagination controls to the top in addition to the bottom would be helpful. See https://datatables.net/examples/advanced_init/dom_multiple_elements.html for an example

Initial rendering of the column page is still slow due to how datatables works. It still has to render all of the rows and then hide them. So, this might not fully solve #162

### DIFF
--- a/src/main/resources/layout/constraint.html
+++ b/src/main/resources/layout/constraint.html
@@ -96,6 +96,6 @@
       </section>
 	  <script>
 	      var config = {
-		      paggination: {{paginationEnabled}}
+		      pagination: {{paginationEnabled}}
 	      }
 	  </script>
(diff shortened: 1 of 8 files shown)


src/main/resources/layout/constraint.html
```diff
@@ -93,4 +93,9 @@ <h3 class="box-title">Check Constraints</h3>
 				</table>				
 			</div>
 		</div>			
-      </section>
\ No newline at end of file
+      </section>
+	  <script>
+	      var config = {
+		      paggination: {{paginationEnabled}}
```
danwatt: This is a misspelling, though I copied it directly from `main.html`

rafalkasa:
@danwatt thank you for your time and contribution to the SchemaSpy project.
Your work is very appreciated. I believe that many people will be happy with this improvement.

PR #171, opened 2017-11-03
Add some test coverage

danwatt:
Add test coverage to:
* Revision.java (not fully covered)
* CaseInsensitiveMap.java
* NameValidator.java
* MustacheWriter.java
* Inflection.java

Also:
* Use try-with-resources in Revision.java
* remove unused constructors from LineWriter
* Use for with Entry instead of while+iterator in CaseInsensitiveMap


src/main/java/org/schemaspy/Revision.java
```diff
@@ -35,33 +37,21 @@
...
-        } catch (IOException exc) {
-        } finally {
-            try {
-                if (reader != null) {
-                    reader.close();
-                } else if (in != null) {
-                    in.close();
-                }
-            } catch (IOException ignore) {}
+        } catch (IOException ignored) {
```
rafalkasa: @danwatt please not leave this empty `catch ` section could you add something like:

```
logger.log(Level.SEVERE, ignored.getStackTrace().toString());
throw IOException;
```

danwatt:
@rafalkasa updated - exception is now logged.

src/main/java/org/schemaspy/Revision.java
```diff
@@ -24,11 +24,14 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.logging.Level;
```
npetzall: Use slf4j

npetzall:
Does it really need a nested try?

danwatt:
@rafalkasa it did in its previous state - `getResourceAsStream` can return null, and `BufferedReader`'s constructor doesnt verify that it got a non-null stream as a parameter. Updating to catch `Exception` will handle this. 

I switched to using the Java provided Manifest parser, as it simplifies the code even more.

src/main/java/org/schemaspy/Revision.java
danwatt: Per request, I've changed to slf4j, but there are no other classes using slf4j. There are 28 classes that import `java.util.logging.Logger`

npetzall:
Really nice, just remove the buildNumber.properties.

danwatt:
Fixed

rafalkasa:
@danwatt now the code looks beautiful 😄 

Maybe the buildNumber.properties file should be ignored or something otherwise always will be committed. Any idea ? @danwatt thanks for already fix this 👍

Issue #145, opened 2017-09-18
Migrate HtmlRoutinesPage class to use MustacheWriter


Riggs333:

danwatt:
Im taking a look at this one

npetzall:
@Riggs333 @danwatt @rafalkasa this can be closed right?

Riggs333:
Sorry for my late response. Yes, looks like this is fixed by now.

PR #179, opened 2017-11-10
Convert Routines to Mustache

danwatt:
This is for #145

I added a table page to list all of the routines, and sub-pages for each routine.

I mostly use SqlServer right now, so I also added support for routines to the base SqlServer configuration. I have a note that needs review on how to proceed with this, as it uses `information_schema`, and this is only supported by SQL Server 2012 and newer

image
image

### DIFF
--- a/src/main/java/org/schemaspy/SchemaAnalyzer.java
+++ b/src/main/java/org/schemaspy/SchemaAnalyzer.java
@@ -211,7 +211,6 @@ public class SchemaAnalyzer {
 
             schemaMeta = null; // done with it so let GC reclaim it
 
-            LineWriter out;
             Collection<Table> tables = new ArrayList<Table>(db.getTables());
             tables.addAll(db.getViews());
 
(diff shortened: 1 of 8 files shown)


src/main/resources/org/schemaspy/types/mssql.properties
```diff
@@ -38,3 +38,7 @@ FROM sysobjects cnstrs \
 INNER JOIN sysobjects tbls ON cnstrs.parent_obj = tbls.id \
 INNER JOIN syscomments cmts ON cmts.id = cnstrs.id \
 WHERE cnstrs.type = 'C' 
+
+selectRoutinesSql=SELECT routine_name, routine_type, data_type AS dtd_identifier, routine_body, routine_definition, is_deterministic, sql_data_access, NULL AS security_type, NULL AS sql_mode, NULL AS routine_comment FROM information_schema.routines WHERE routine_schema = :schema
+
+selectRoutineParametersSql=SELECT specific_name, parameter_name, data_type AS dtd_identifier, parameter_mode FROM information_schema.parameters WHERE specific_schema = :schema AND ordinal_position != 0 ORDER BY ordinal_position
```
danwatt: `information_schema` is not supported prior to SqlServer 2012.

Should a new properties file be added for 2012? (this would require 3 - one for the sql driver, and two for jtds)

rafalkasa:
Amazing work I'm impressed ! Thank you for all efforts for the SchemaSpy !

danwatt:
@rafalkasa - I think I should separate the SQLServer changes out into a separate PR. It was useful for testing, but I think there is some discussion that needs to happen on how to support routines on SQLServer - either another set of properties files, or raise the minimum supported versions of SQLServer

rafalkasa:
This routine enhancement is something that many people will loved ! We have to add the databse configuration properties `selectRoutinesSql` and `selectRoutineParametersSql` for the top used databases like Oracle, MSSQL and PostgreSQL. I have checked and these configuration exists for MySql and MariaDB

PR #182, opened 2017-11-13
Support Routines in SQLServer

danwatt:
This will only work on SQLServer 2012 and later.

Discussion - should we create a separate properties file for `mssql2012`, or just raise the baseline?

### DIFF
--- a/src/main/resources/org/schemaspy/types/mssql.properties
+++ b/src/main/resources/org/schemaspy/types/mssql.properties
@@ -38,3 +38,12 @@ FROM sysobjects cnstrs \
 INNER JOIN sysobjects tbls ON cnstrs.parent_obj = tbls.id \
 INNER JOIN syscomments cmts ON cmts.id = cnstrs.id \
 WHERE cnstrs.type = 'C' 
+
+selectRoutinesSql=SELECT routine_name, routine_type, data_type AS dtd_identifier, routine_body, routine_definition, is_deterministic, sql_data_access, \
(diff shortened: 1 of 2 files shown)


npetzall:
I say create a new file.
Motivation: People that really need schemaspy are the ones running legacy systems and they might be leeegacy

npetzall:
Just to check my self.
Didn't this just become a bit strange?

Rutines that are available in 2012 are added in mssql which by the looks of it is for SQL 2000, the runtines queries are nulled in mssql05 which by comment is for SQL 2005+.

On a totally different note, should the routines queries be added to the jdts dbtype?

danwatt:
Yea, @rafalkasa , based on the one comment that @npetzall , I think this should be in a separate file. I can create a new PR to do that.

So, we will have support for three different versions of sqlserver : 2000+, 2005+, and 2012+

rafalkasa:
@danwatt, @npetzall grate, but I would like to change the names of the db types files to:

**Option 1:**
- `mssql.properties` - 2012+
- `mssql05.properties` - 2005, 2008
- `mssql00.properties` - 2000

**Option 2:**
- `mssql12.properties` - 2012+
- `mssql05.properties` - 2005, 2008
- `mssql.properties` - 2000
 
Which option do you would like to choose ?

danwatt:
Option 2

rafalkasa:
@npetzall thank you for explanation and arguments. I also think that for backward compatibility we should choose the Option 2.

PR #186, opened 2017-11-20
Handle when routine definition is null. (For instance in SQL Server w…

martenbohlin:
…hen a rutine is defined in an external language)

In SQL server it is not always possible to query the definition of a routine. This patch handles that case.

### DIFF
--- a/src/main/java/org/schemaspy/model/Routine.java
+++ b/src/main/java/org/schemaspy/model/Routine.java
@@ -60,7 +60,7 @@ public class Routine implements Comparable<Routine> {
         this.type = type;
         this.returnType = returnType;
         this.definitionLanguage = definitionLanguage;
-        this.definition = definition;
+        this.definition = coalesce(definition, "");
         this.dataAccess = dataAccess;
         this.securityType = securityType;
         this.deterministic = deterministic;
@@ -159,5 +159,9 @@ public class Routine implements Comparable<Routine> {
         if (rc == 0)
             rc = getDefinition().compareTo(other.getDefinition());
         return rc;
+    }
+
+    private <T> T coalesce(T a, T b) {
+        return a!=null ? a : b;
     }
 }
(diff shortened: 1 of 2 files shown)


npetzall:
I made a quick check and getDefinition is used twice.
HtmlRoutinesPage and in Routine.compareTo(Routine other)

This fix fixes HtmlRoutinesPage but the compareTo method currently doesn't allow definition to be null due to the syntax

```
rc = getDefinition().compareTo(other.getDefinition());
``` 

So I'm thinking either default to empty string in Routines when created, either in the class or the creating class/object. Alternative solution would be to check if compareTo is actually used and remove if not used.

npetzall:
Just remembered check #182 it's closed but the dbtype mssql supports routines but mssql05 doesn't.

martenbohlin:
Good catch! I have moved the null check to the constructor, since that felt as the most safe option. (It also avoids future problems).

Should I keep the pullrequests as two commits, or is it better to squash them together?

I have looked at #182, and I do not see any conflicts with that change.

npetzall:
Squash or fix up, should be one commit.

#182 wasn't conflict. Previously mssql was for SQL 2000 to 2005 and mssql05 was 2005 and newer with #182 mssql became 2012 and newer. 

So you might get routines populated with the mssql dbtype instead of mssql05 if you're using that and have routines in the database.

rafalkasa:
@martenbohlin could you provide unit tests for your changes. Thank you in advance.

martenbohlin:
Test added

npetzall:
By no means anything you need to fix, but have you tried assertj? Just a question.

martenbohlin:
* Changed to assertj since that is what is mostly used in the project.
* Include code and test in same commit

rafalkasa:
@martenbohlin thank you for the unit tests whole contribution looks awesome and is ready to merge 😄

rafalkasa:
@martenbohlin  check your email used for commit contribution on github account.
Maybe this link wilk help you https://help.github.com/articles/why-are-my-contributions-not-showing-up-on-my-profile/

martenbohlin:
Thanks! I will check on my email setting for the next pull request.

rafalkasa:
You have to add email that you used for commits to the Github account. You can do this any time

martenbohlin:
Aha thanks. Thought I had to change the one used in my commits.

PR #187, opened 2017-11-22
Move Routines support for SQLServer into new type - mssql12

danwatt:
Per our discussion in #182, this moves the support for querying routines to a new database type - mssql12.

This also adds support for routines to the JDTS driver

* mssql12 extends mssql05 extends mssql
* mssql12-jtds extends mssql05-jtds extends mssql-jtds
* mssql12-jtds-instance extends mssql05-jtds-instance extends mssql-jtds-instance

### DIFF
--- a/src/main/resources/org/schemaspy/types/mssql.properties
+++ b/src/main/resources/org/schemaspy/types/mssql.properties
@@ -37,13 +37,4 @@ cmts.text AS text \
 FROM sysobjects cnstrs \
 INNER JOIN sysobjects tbls ON cnstrs.parent_obj = tbls.id \
 INNER JOIN syscomments cmts ON cmts.id = cnstrs.id \
-WHERE cnstrs.type = 'C' 
-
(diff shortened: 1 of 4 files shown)


rafalkasa:
@danwatt thank you for your contribution I'm more then happy and also other SchemaSpy users will be.
Before we will release the new version I will prepare the documentation for that.

rafalkasa:
@danwatt I already testing the routines support change on MSSQL 2008 R2 and is working.
What mean that routines query used by you working not only for MSSQL 2012 but also for lower versions of SQL Server. In this situation maybe better will be check if the same code is also supported in MSSQL 2005. Do you have possibility to test ?

rafalkasa:
I verified this on Microsoft site they added `information_schema.routines` from 
MSSQL 2008 version: https://docs.microsoft.com/en-us/sql/relational-databases/system-information-schema-views/routines-transact-sql

danwatt:
If it is working on 2008, then yes, it should be changed to 2008. I don't have access to 2008, only 2012

rafalkasa:
I tested today and yesterday on 2008 and is working also in official documentation was mentioned about it.

PR #189, opened 2017-11-29
Clean up Sonar warnings around File deletion

danwatt:
There are several Sonar warnings around "Do something with the "boolean" value returned by "delete"".

HtmlTablePage.java in particular attempts to delete all of the files that will be generated, but doesn't check to see if they were actually deleted. Its possible that the files could be locked by another process.

A couple different changes were made, trying to keep with the intent of the original code while providing better error handling.

* [FileUtils.forceMkdir](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/FileUtils.java#L2474) - Will throw an exception if the input already exists but is not a directory, otherwise attempts to call mkdirs. If that fails, throws an IOException
* Files.deleteIfExists - deletes a file if it exists, throws an IOException if it exists and can't be deleted (ie: the file is locked by another process)
* FileUtils.deleteQuietly - added in a couple of exception handlers

### DIFF
--- a/src/main/java/org/schemaspy/SchemaAnalyzer.java
+++ b/src/main/java/org/schemaspy/SchemaAnalyzer.java
@@ -18,6 +18,7 @@
  */
 package org.schemaspy;
 
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.filefilter.FileFilterUtils;
 import org.apache.commons.io.filefilter.IOFileFilter;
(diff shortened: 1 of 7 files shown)


npetzall:
Looks good, don't see any apparent problems.

Saw some stuff that should be refactored but that's another story.

danwatt:
Yea, trying to keep refactors to a separate PR. This was purely to clean up warnings.

PR #191, opened 2017-11-29
Switch to SLF4J

danwatt:
* use `LOGGER` for consistency
* setting the logging level from the CLI has been removed, as SFL4J itself does not expose a method to do this
* fineEnabled / finerEnabled / etc... checks have been removed
* Use SLF4J's formatting capabilities to remove a lot of warnings about doing String concatenation on logging calls


rafalkasa:
@danwatt Wow you are the Titan of work, awesome thanks.

src/main/java/org/schemaspy/service/TableService.java
```diff
@@ -118,8 +116,8 @@ private void initColumnAutoUpdate(Database db, Table table, boolean forceQuotes)
             if (forceQuotes) {
                 if (!table.isLogical()) {
                     // don't completely choke just because we couldn't do this....
-                    LOGGER.warning("Failed to determine auto increment status: " + exc);
-                    LOGGER.warning("SQL: " + sql.toString());
+                    LOGGER.warn("Failed to determine auto increment status: {}", exc);
```
npetzall: I just saw this.
I'm thinking and this is my subjective view.

One should never call logging on the same level twice, for one error.
In our case, there might not be a concurrency issue, but as a rule of thumb.

This is to avoid:
warn - Failed to deter....
Bunch of other loggs from other threads
warn - SQL: ....

This should probably be refactored into:
```
LOGGER.warn("Failed to determine auto increment status using SQL: {}", sql.toString(), exc);
```

npetzall:
```
JUL -> SLF4J
FINEST  -> TRACE
FINER   -> DEBUG
FINE    -> DEBUG
INFO    -> INFO
WARNING -> WARN
SEVERE  -> ERROR
```
https://www.slf4j.org/apidocs/org/slf4j/bridge/SLF4JBridgeHandler.html

Some of the finer are mapped to trace instead of debug.

npetzall:
Looks good to me. And the conflicts are easy.

rafalkasa:
@danwatt please resolve the conflicts and I will merge the code

danwatt:
@rafalkasa merge conflicts resolved.

src/main/java/org/schemaspy/Config.java
```diff
@@ -1220,57 +1221,6 @@ public boolean isLowQuality() {
...
-     * <li><code>info</code>
-     * <li><code>config</code>
-     * <li><code>fine</code>
-     * <li><code>finer</code>
-     * <li><code>finest</code>  (lowest - most detail)
-     * </ul>
-     *
-     * @param logLevel
-     */
-    public void setLogLevel(String logLevel) {
```
rafalkasa: @danwatt after when we delete setLogLevel parameter is not possible to change application.properties/logging.level.org.schemaspy from command line. 

This was common feature in previous version of SchemaSpy in many cases original error was hidden as long as -loglevel param was not set to higher.

When we will run the SchemaSpy with `java -jar schemaspy.jar --debug` param we will only change `logging.level.root` but not `logging.level.org.schemaspy`

npetzall: I know this is a bit old but...
`--logging.level.org.schemaspy=DEBUG`
works.

Possible solutions:
1, We could use spring placeholder and make it `--loglevel` still the double dash.
2, We could add -loglevel to jcommander and have it injected in to the spring environment.

npetzall: Could just do a `-debug` i Jcommander.

PR #196, opened 2017-12-06
More relaxed handling on missing Database/ResultSetMetadata

npetzall:
* We have an SQL92 keywords list, so we don't need to fail if we can't
  get from database
* If we can't retrieve column from table model using ResulSetMetaData
  check if it contains table/fulltable name in the MetaData and strip it.
  If it's still not found throw an InconsistencyException (this is for
  when DatabaseMetaData returns columns in one way and ResultSetMetaData
  in another.
* Added basic hive dbtype connectionSpec/connectionString has a bunch of
  options: https://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.3.0/bk_dataintegration/content/hive-jdbc-odbc-drivers.html


rafalkasa:
For me the changes looks good I will merge this branch with master

Issue #134, opened 2017-08-16
Failed to determine auto increment status when database has spaces in name


bvukas:
In MySQL I have database named `Test 1.0` and table named `Table 1.0`. 

This is log output:

Aug 16, 2017 4:29:53 PM org.springframework.context.annotation.AnnotationConfigApplicationContext prepareRefresh
INFO: Refreshing org.springframework.context.annotation.AnnotationConfigApplicationContext@30dae81: startup date [Wed Aug 16 16:29:53 CEST 2017]; root of context hierarchy
Aug 16, 2017 4:29:53 PM org.schemaspy.Config loadProperties
INFO: Configuration file not found
Using database properties:
  [schemaspy-6.0.0-rc1.jar]/org/schemaspy/types/mysql.properties
Gathering schema details...16:29:55.162 WARNING: TableService.initColumnAutoUpdate - Failed to determine auto increment status: java.sql.SQLException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '1.0.`Table 1.0` where 0 = 1' at line 1
16:29:55.163 WARNING: TableService.initColumnAutoUpdate - SQL: select * from Test 1.0.`Table 1.0` where 0 = 1
.(0sec)
Connecting relationships....(0sec)
Writing/graphing summary..........(0sec)
Writing/diagramming details.Table -> Table 1.0
(0sec)
Wrote relationship details of 1 tables/views to directory './Test10' in 2 seconds.
View the results by opening ./Test10/index.html

As result, id column in Table 1.0 is not marked as auto incremented

npetzall:
Do you get correct output except the auto increment on id column.

bvukas:
Honestly, I did not check everything else, but I did not notice that anything else was broken.

npetzall:
I just asked since I wanted to scope down the issue. To possibly one method.

But if it's just a db called "Test 1.0", a table called "Test 1.0" and a Pk column id with autoincrement it shouldn't be that hard to reproduce.

rafalkasa:
This issue is quite interesting I don't imagine before that somebody can have table or database with space in the name. Thank you @bvukas for report the issue.

bvukas:
Script to create database and table that can be used to reproduce issue:

/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
[...]

PR #201, opened 2017-12-16
Removed getQuotedIdentifier for tableName when using MetaData.getColumns

npetzall:
* Removing since it was added due to bug in mysql-connector-java
  https://bugs.mysql.com/bug.php?id=63992

* Better output in ResultSetDumper.dumpAll
* Added script function for H2MemoryRule
* Added lots of tests
* Ignoring MysqlSpacesIT test and using MysqlSpacesNoDotsIT


npetzall:
So spaces works but not dots, I've bumped the bug at mysql.
And added comments to the closed issue #134

PR #202, opened 2017-12-18
Render "dot" files without graphviz native code. Activated by CLI parameter "-dotinjvm true"

igor-suhorukov:
This changes are render graphviz "dot" files in JVM only by using https://github.com/mdaines/viz.js/

JDot mode activated by "-dotinjvm true" command line parameter, when external "dot" command is not available


npetzall:
Cool, I was looking into this prior to some other stuff. Viz.js support images so, no need to replace them with text.

npetzall:
I'm not 100% on this but I think the option should be added to Commandlinearguments instead of Config. But that can be done later actually.

igor-suhorukov:
@npetzall I try to use image in Viz.js but without luck

igor-suhorukov:
> I'm not 100% on this but I think the option should be added to Commandlinearguments instead of Config. 

Thanks @npetzall  I will move this option to Commandlinearguments

npetzall:
Sadly it doesn't work right of the bat, you need to specify image path, and size in the call to viz as an option images.

I've added support to d3-graphviz but that's pure JavaScript thing.

src/main/java/org/schemaspy/util/JDot.java
```diff
@@ -0,0 +1,62 @@
...
+        } catch (Exception e) {
+            throw new IllegalArgumentException("viz.js", e);
+        }
+    }
+
+    protected String toSvg(String dotSource, int memSize) {
+        try {
+            scriptEngine.getBindings(ScriptContext.ENGINE_SCOPE).put("dotSource", dotSource);
+            scriptEngine.getBindings(ScriptContext.ENGINE_SCOPE).put("memSize", memSize);
+            return (String) scriptEngine.eval("Viz(dotSource,{ totalMemory: memSize })");
```
npetzall:
```
return (String) scriptEngine.eval("Viz(dotSource, { totalMemory: memSize, images: [ { path: "images/primaryKeys.png" , width: "261px" , height: "261px" },{ path: "images/foreignKeys.png" , width: "261px" , height: "261px" }] })");
```

Something like that

igor-suhorukov: @npetzall  thanks for assisting! Could you please review latest code changes after your comments?

igor-suhorukov:
@npetzall sorry) I will fix broken integration tests for this pull request

igor-suhorukov:
@npetzall pull request ready for code review

src/main/java/org/schemaspy/util/Dot.java
```diff
@@ -317,7 +332,7 @@ public String generateDiagram(File dotFile, File diagramFile) throws DotFailure
         }
     }
 
-    public class DotFailure extends IOException {
+    public static class DotFailure extends IOException {
```
npetzall: Extract to own class.

npetzall:
Some tests would be nice.

src/main/java/org/schemaspy/util/JDot.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public String getVersion(){
+        return SIMULATE_VERSION;
+    }
+
+    public String renderDotByJvm(File dotFile, File diagramFile) throws Dot.DotFailure {
+        try {
+            String dotSource = IOUtils.toString(dotFile.toURI().toURL(), "UTF-8");
+            String svg = toSvg(dotSource, 1 << 30);
```
npetzall: Extract `1 << 30` to a static variable to explain the value.

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -76,6 +71,9 @@ public SchemaAnalyzer(SqlService sqlService, DatabaseService databaseService, Co
         this.databaseService = Objects.requireNonNull(databaseService);
         this.commandLineArguments = Objects.requireNonNull(commandLineArguments);
         addOutputProducer(new XmlProducerUsingDOM());
+        if(commandLineArguments.isRenderDotInJvm()){
+            Dot.setInstance(commandLineArguments.isRenderDotInJvm(), commandLineArguments.getOutputDirectory().getName());
```
npetzall: Wonder if this could be put in the SchemaSpyConfiguration instead.
It would be a step in the right direction regarding dependency injection.
Possibly due to the current state instead of setInstance have a getOrCreateInstance() and have it return that instance. It would then be present in the SpringContext.

But it can be done later.

igor-suhorukov:
@npetzall thank you very much for code review and help with images. I really appreciate your work!

src/main/java/org/schemaspy/util/JDot.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.script.ScriptException;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.InputStream;
+
+public class JDot {
+
+    private static final String SIMULATE_VERSION = "2.26";
+    private static final String ICON_SIZE = " , width: \"261px\" , height: \"261px\"";
+    private static final int JAVA_SCRIPT_ENGINE_MEMORY_SIZE = 1 << 30;
```
npetzall: I was thinking more like ONE_GB.

The usage location will tell you what's the variable is used for but a human readable explanation for the value would be ONE_GB.

igor-suhorukov:
@npetzall done

npetzall:
@rafalkasa I think this looks awesome as a first step.

I actually looked deeper into graphviz-java and it always goes thur svg before png so any bad quality would be present in graphviz-java. As I said it's a good first step.

What we do later will be later.

rafalkasa:
@igor-suhorukov, @npetzall thank you for your work and cooperation, I need spent some time for review the changes in PR and make some additional tests. 

I also believe that we have to do something to be not dependent to the graphvize external library. Your work is the first big step to improve this part of the project.

rafalkasa:
@igor-suhorukov could you check this error:

```
.org.schemaspy.util.DotFailure: javax.script.ScriptException: Error: No or improper image file="C:\develop\source_database\docs/images/primaryKeys.png"
 in <eval> at line number 94 at column number 4
```
I'm working on Windows and in my opinion this is linux/windows path specyfic issue.
Only differences that I noticed was that graphviz has better performance speed.

igor-suhorukov:
thanks @rafalkasa for review and bug report. I fix it

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -76,6 +71,9 @@ public SchemaAnalyzer(SqlService sqlService, DatabaseService databaseService, Co
         this.databaseService = Objects.requireNonNull(databaseService);
         this.commandLineArguments = Objects.requireNonNull(commandLineArguments);
         addOutputProducer(new XmlProducerUsingDOM());
+        if(commandLineArguments.isRenderDotInJvm()){
```
npetzall: Missing else statement, but as I commented earlier maybe move this to Schemaspyconfiguration.

rafalkasa:
@igor-suhorukov I don't know where is the issue but the error:

```
....org.schemaspy.util.DotFailure: javax.script.ScriptException: Error: No or improper image file="C:\develop\insure\docs\database/images/primaryKeys.png"
 in <eval> at line number 94 at column number 4
```

is related to the line: 
https://github.com/schemaspy/schemaspy/blob/e78ab2433691e9fc28e18d1912d0bd5290ae3b4a/src/main/java/org/schemaspy/view/DotNode.java#L139

I try to change the path using `FilenameUtils.separatorsToSystem` and `FilenameUtils.separatorsToUnix` but after the changes error stay the same... 
Did you test this on your environment and icons generate correct without any error ? Maybe I have to verify how this will be working on linux machine.

rafalkasa:
@igor-suhorukov thank you for your update, but nothing change I have still the same error.
How this is working for you ? What OS do you use ?

npetzall:
There might be two issues.
1, viz can't actually read the file
2, the path in images option isn't the same as the path in the dot file.

I'm guessing on the latter, but it looked correct when checking the source. Probably need to debug to find out if one might be absolut and the other one relative. That might cause issues.

igor-suhorukov:
@rafalkasa I use Linux system to develop. I will try run it on Windows. Root cause of issue that I test in on linux with relative path and not try to use absolute path

igor-suhorukov:
@npetzall thanks I will try to debug pull request on different output paths and in Win OS famaly

igor-suhorukov:
Hi @rafalkasa @npetzall ! vis.js just unsupport backslashes in path.
New code version should work in *nix and windows environments now

igor-suhorukov:
@rafalkasa I test last PR change on Windows 10. All works fine

igor-suhorukov:
Hi @rafalkasa 
Any updates?

rafalkasa:
@igor-suhorukov right now diagrams is generated correctly, but instead of icon represent the primary key or foreign key missing graphic is generated.

image

Could you check by you how this generated diagram looks on your computer ?

igor-suhorukov:
Hi @rafalkasa 
It depends on view mode: 

SVG embedded in html 
screen-1

The same SVG in new window:
screen-2

igor-suhorukov:
I can convert SVG to PNG on [java side](https://xmlgraphics.apache.org/batik/using/transcoder.html)

rafalkasa:
@igor-suhorukov yes I checked and it's as you wrote, hmm please try to do something to show correct looking diagrams. Maybe instead converting svg to png is other option to present generated diagrams correctly

npetzall:
Please look at the img element in the svg. It might have a relative path which is good in a sense but bad since the opening the svg has one relative path and opening "embedded" will have another path.

This could be handled in relationship.js so that the path is updated.

rafalkasa:
Another solution is to use an `embed` tag instead of an `img` tag for the picture.
`<embed src="svg.svg">` I have tested this second ago and is working as a charm 😄

npetzall:
@rafalkasa is it possible to move the whole output folder and it still works?

rafalkasa:
@igor-suhorukov I have done some additional verification. Today or tomorrow I will share with you what we should to improve. Please in the meantime fix the conflicts.

igor-suhorukov:
@rafalkasa thank you for feedback. I resolve merge conflict

npetzall:
@igor-suhorukov 
Could you change org/schemaspy/testcontainer/MysqlHTMLIT.java:112
from
```
return a.compareTo(e);
```
to
```
return a.trim().compareTo(trimmed);
```
It should make your PR pass.

I added a full html test and it's brittle as hell. This will add some robustness to it.

igor-suhorukov:
thank you @npetzall

npetzall:
Well I feel responsible so I just want to say that I'm sorry.

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -73,6 +73,8 @@ public SchemaAnalyzer(SqlService sqlService, DatabaseService databaseService, Co
         this.databaseService = Objects.requireNonNull(databaseService);
         this.commandLineArguments = Objects.requireNonNull(commandLineArguments);
         addOutputProducer(new XmlProducerUsingDOM());
+        File outputDirectory = commandLineArguments.getOutputDirectory();
+        Dot.setInstance(commandLineArguments.isRenderDotInJvm(), outputDirectory!=null?outputDirectory.getPath():null);
```
npetzall: outputDirectory is a required parameter you shouldn't need check it for null
But since we bring the spring context up even in cases when we aren't going to use outputDir such as -help -dbhelp you'll need the null check.

I'm gonna think about that, would be nice to no require null checks on required parameters.

igor-suhorukov: Too many tests failed without this check:
[ERROR] Tests run: 37, Failures: 0, Errors: 25, Skipped: 6

npetzall: Yeah, figured it. Only way to avoid it would be to use output Directory later.

So that the getPath is only called if it's actually going to be used.

Since all constructors will be called, during spring context startup even if they won't be used.
One might argue that dot stuff shouldn't be initialized if -nohtml. And the 37 test that fail should have -nohtml.

npetzall:
I did some playing around today with viz.js and it's actually possible but not documented that you can get <map> information from it. You can actually get PNG as base64 and that's documented. So we could do the old png and <map> with viz.js

But the svg quality by viz.js is way better than the svg that is hosted in the example http://schemaspy.org/sample/diagrams/Album.1degree.svg

So the only fear I have is the possible memory issue when running huge databases else I would want to only use viz.js

npetzall:
As a follow up to my previous comment. If we think we can go full viz.js we should not write dot-file and we should not write .svg files but just append the svg element directly into the DOM. This would allow us to manipulate the SVG (zoom, pan, highlight, etc, etc, hide, collapse, move)

src/main/java/org/schemaspy/util/DotFailure.java
```diff
@@ -0,0 +1,11 @@
+package org.schemaspy.util;
```
npetzall: Missing copyright and license

src/main/java/org/schemaspy/util/JDot.java
```diff
@@ -0,0 +1,68 @@
+package org.schemaspy.util;
```
npetzall: Missing copyright and license

rafalkasa:
Code was merged with #456 and we are working on release this feature with version 6.1.0 of SchemaSpy @igor-suhorukov thank you for your contribution.

rafalkasa:
I close this PR changes already merged to the master by #456

piaste:
The ReadTheDocs site doesn't mention this new feature.

As a user, what reasons should I have to choose viz.js over the default native graphvizDot, or vice versa?

npetzall:
@piaste
https://schemaspy.readthedocs.io/en/v6.1.0/new.html
https://schemaspy.readthedocs.io/en/v6.1.0/configuration/commandline.html at the end. https://schemaspy.readthedocs.io/en/v6.1.0/installation html optional heading.

piaste:
🤦‍♂ Thanks, I missed that the PR was different from what merged and I had searched for 'dotinvjm'.

I'm using schemaspy through Docker so I guess I'll keep using the native version since it's included in the image.

npetzall:
Yepp, use the native. I think it's a bit faster and I've done some tests with running multiple dot instances so there is a potential to have it work faster. Have tested multithreading vizjs yet.

PR #208, opened 2017-12-20
More cleanups

danwatt:
Most of these fix some sonar warnings
* Replace most printStackTrace() calls to use a logger - a few remain, they are grouped with a bunch of other System.out lines
* Remove some dead stores that were put in there a long time ago to "allow GC ASAP"
* Several classes had compareTo methods, but no equals / hashcode
* Switched a manual null check in a constructor to use `Objects.requireNonNull`
* There were a few uses of string concatenation inside of a loop, which isn't as efficient as allocating a StringBuilder (sonar  squid:S1643)

Fix the following IntelliJ warnings:
* for loop replaceable with foreach
* indexOf expression is replaceable with contains
* unnecessary boxing
* unnecessary unboxing
* Collections.sort could be replaced with List.sort
* Lambda can be replaced with method reference
* Duplicate throws

### DIFF
--- a/src/main/java/org/schemaspy/model/Catalog.java
+++ b/src/main/java/org/schemaspy/model/Catalog.java
@@ -24,28 +24,23 @@ import java.util.Objects;
 import org.schemaspy.util.CaseInsensitiveMap;
 
 public final class Catalog implements Comparable<Catalog>{
-    private String name;
-    private String comment = null;
+    private final String name;
+    private String comment;
 
 	public Catalog(String name) {
-		super();
-		Objects.requireNonNull(name);
-		this.name = name;
+		this(name,null);
+
 	}
 	
 	public Catalog(String name, String comment) {
-		this(name);
-		this.name = name;
+		this.name = Objects.requireNonNull(name);
 		this.comment = comment;
 	}
 	
 	public String getName() {
 		return name;
 	}
-	public void setName(String name) {
-		this.name = name;
-	}
-	
+
 	public String getComment() {
 		return comment;
 	}
@@ -67,7 +62,6 @@ public final class Catalog implements Comparable<Catalog>{
 	}
 
 	@Override public int hashCode() {
-
-		return Objects.hash(name);
+		return name.hashCode();
 	}
 }
--- a/src/main/java/org/schemaspy/model/Catalog.java
+++ b/src/main/java/org/schemaspy/model/Catalog.java
@@ -18,10 +18,7 @@
  */
 package org.schemaspy.model;
 
-import java.util.Map;
 import java.util.Objects;
-
-import org.schemaspy.util.CaseInsensitiveMap;
 
 public final class Catalog implements Comparable<Catalog>{
     private final String name;
--- a/src/main/java/org/schemaspy/model/Catalog.java
+++ b/src/main/java/org/schemaspy/model/Catalog.java
@@ -19,25 +19,22 @@
 package org.schemaspy.model;
 
 import java.util.Map;
+import java.util.Objects;
 
 import org.schemaspy.util.CaseInsensitiveMap;
 
 public final class Catalog implements Comparable<Catalog>{
     private String name;
     private String comment = null;
-    private Map<String, Schema> schemas = new CaseInsensitiveMap<Schema>();
-    
+
 	public Catalog(String name) {
 		super();
-		 if (name == null)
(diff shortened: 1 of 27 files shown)


src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -19,25 +19,22 @@
 package org.schemaspy.model;
 
 import java.util.Map;
+import java.util.Objects;
 
 import org.schemaspy.util.CaseInsensitiveMap;
 
 public final class Catalog implements Comparable<Catalog>{
     private String name;
```
Riggs333: Consider making the field `name` final and remove the setter as it is never used (as far as I can see).

src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -19,25 +19,22 @@
...
     private String comment = null;
-    private Map<String, Schema> schemas = new CaseInsensitiveMap<Schema>();
-    
+
 	public Catalog(String name) {
 		super();
```
Riggs333: Remove the unnecessary `super()` call.

src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -19,25 +19,22 @@
...
-		 if (name == null)
-	            throw new IllegalArgumentException("Catalog name can't be null");
+		Objects.requireNonNull(name);
 		this.name = name;
```
Riggs333: It's a matter of taste but as `Objects#requireNonNull` returns the given value (unless it's `null) we can join the lines 32 and 33.

src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -19,25 +19,22 @@
...
 	}
 	
 	public Catalog(String name, String comment) {
-		super();
-		 if (name == null)
-	            throw new IllegalArgumentException("Catalog name can't be null");
+		this(name);
 		this.name = name;
```
Riggs333: As the one-argument-constructor is called here (`this(name)`) there is no need to assign the value to the field `name` again. So line 38 can be removed.

src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -61,5 +58,16 @@ public int compareTo(Catalog i) {
...
+
+	@Override public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
+		Catalog catalog = (Catalog) o;
+		return Objects.equals(name, catalog.name);
+	}
+
+	@Override public int hashCode() {
+
```
Riggs333: Please remove this empty line.

src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -61,5 +58,16 @@ public int compareTo(Catalog i) {
...
+		return Objects.hash(name);
```
Riggs333: IntelliJ complains on this line about "Object instantiation inside `hashCode()` (varargs call)".
Should not make that much of a difference but I propose that a `name.hashCode()` is enough here. ;-)

src/main/java/org/schemaspy/model/Catalog.java
```diff
@@ -61,5 +58,16 @@ public int compareTo(Catalog i) {
     public String toString() {
         return name;
     }
-    
+
+	@Override public boolean equals(Object o) {
```
Riggs333: Please add test(s) for this new code.
My proposal is to use the brilliant [EqualsVerifier](http://jqno.nl/equalsverifier/) that makes those tests more or less a oneliner.

src/test/java/org/schemaspy/model/CatalogTest.java
```diff
@@ -0,0 +1,29 @@
+package org.schemaspy.model;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CatalogTest {
```
Riggs333: My proposal is to use the brilliant [EqualsVerifier](http://jqno.nl/equalsverifier/) that makes those tests more or less a oneliner.

danwatt:
Thanks @Riggs333 for the suggestions.

There are a lot more changes that could be done to model classes - very inconsistent use of final fields, some do a lot more than just hold values (Database), most lack tests, etc.

PR #209, opened 2017-12-20
Switch MSSQL2012 to MSSQL2008

danwatt:
Per discussion in #187


rafalkasa:
Thank you @danwatt for memory about that

PR #213, opened 2017-12-24
Add export option to Excel, PDF, CSV, Clipboard

rafalkasa:
This pull request contains new functionality allow SchemaSpy users export information about tables, procedures, columns to Excel, PDF, CSV and Clipboard.

User will see new buttons group on each page that allow them very fast and easily export information about the database objects.

image


npetzall:
I'm not a frontend guy. But this is a frontend maven plug in. 
https://github.com/eirslett/frontend-maven-plugin

PR #218, opened 2018-01-14
Adding informix index test

npetzall:
* Added testcontainer for informix
* Added a assumption rule for driverClass that is quicker than
  testcontainer-junit.
* Added ignore to /database/@type for XML comparison.


rafalkasa:
@npetzall if the assumption rule for driverClass is faster than testcontainer-junit maybe it will be good to add this also for added last time MSSQLServerIT ? So you have to merge first wit master branch.

npetzall:
Not needed for mssql or mysql since drivers are defined in pom and available in maven.

rafalkasa:
ok so I will merge your changes

rafalkasa:
Super this is a big difference, congratulations

PR #192, opened 2017-11-30
157 isolating spring to app

npetzall:
Moving CLI/App to own package.
This is a step in making CLI/App it's own module and isolating spring and cli from currently everything else.

PR #234, opened 2018-01-24
Full html test

npetzall:
Fixed html in test.
Added test to compare html and a cpl of lines to ignore.

Related to #192


npetzall:
Most files are just added as expectations to test.

src/test/java/org/schemaspy/testcontainer/MysqlHTMLIT.java
```diff
@@ -0,0 +1,118 @@
...
+                return fileName.endsWith("html") || fileName.endsWith("dot");
+            }).collect(Collectors.toList());
+        }
+        assertThat(expectations.size()).isGreaterThan(0);
+        SoftAssertions softAssertions = new SoftAssertions();
+        for (Path expect : expectations) {
+            List<String> expectLines = Files.readAllLines(expect, StandardCharsets.UTF_8);
+            Path actual = Paths.get(target, expectedPath.relativize(expect).toString());
+            List<String> actualLines = Files.readAllLines(actual, StandardCharsets.UTF_8);
+            softAssertions.assertThat(actualLines).as("%s doesn't have the expected number of lines: %s", actual.toString(), expectLines.size()).hasSameSizeAs(expectLines);
```
rafalkasa: @npetzall I'm afraid that each time when we change HTML layout we must generate the expected HTML files for this test. Otherwise expectLines will be not equal to actualLines and this test will never pass... ? Correct me if I'm wrong.

npetzall: Yepp, you are correct, we can have it disabled. I only created it to be able to do refactoring of mustachwriter. Validate that my changes didn't have any side-effects.

rafalkasa: @npetzall I will conditionally merge your PR as it is to the master, but if we will have the problem with `producesSameContet` test in the future I will skip this test.

npetzall: Might be good idea to disable after #237 I can look into some kind of scripto update the expected.

Issue #157, opened 2017-10-09
Maven Central


leoherbie:
Please release to maven central.

npetzall:
Just to check. Schemaspy isn't really a library but an application so putting it on maven doesn't really add any benefits. Url to download from Maven vs url to download from github.

Unless you want to use it as a library if that is the intention, then we should probably refactor the code and make it in to a cpl of libraries and an application.

rafalkasa:
@npetzall many users use SchemaSpy as Maven tool to lunch after the build generation process of database documentation.

Maven is not only used as library dependencies, this is how I understand @leoherbie proposition.

npetzall:
I only ask since there might be areas we can improve to support some usecase better. Like using it in build process to generate up-to-date database documentation. It might be good to have it as a maven plugin that can publish to maven site.

rafalkasa:
Good point! I would like to do with SchemaSpy something similar to the flyway https://flywaydb.org/getstarted/firststeps/maven

They have core library, API, command line tool and maven plugin.
I agree with you that it should be planed. Thank you for your opinion and suggestion.

leoherbie:
I'm glad this issue caused a bit of discussion.  For us, we embed the classic schemaspy in our application as a library.  Then on startup we generate the schemaspy metadata and provide links to the generated artifacts from an admin section of our application.  One consideration is the old schemaspy did not have any third party dependencies so it worked great as a dependency.  I see the new schemaspy is a spring boot app.  This may make it a bit harder to include.

I could also see someone using it during a build phase as well even though this is not our usecase

leoherbie:
BTW.  A minor consideration for schemaspy as a lib usecase I just thought of.  The old progress bar sysout output is not very friendly for applications that use a logging framework.  We thought about fixing this in our schemaspy 5 fork but never did.  You just get this constant ........ output continually printing.  Not really related to maven but is related to making it library friendly.

npetzall:
Progress bar is somewhat reported in #119.

The "core" library and it's dependencies is going to be a bit trickier.

@leoherbie So for a person wanting to run your application they need graphviz installed?

leoherbie:
Yes.  They do indeed need graphviz installed although as I recall it can still work partially without it.

npetzall:
I propose starting by separating "functions" into packages.

Example 
org.schemaspy.db: anything related to extracting data from db
org.schemaspy.output: anything generated by schemaspy
[...]

PR #235, opened 2018-01-24
Removed stereotype from all classes not tied to application

npetzall:
* Added new wiring configuration
* Left stereotypes in commandline classes

### DIFF
--- a/src/main/java/org/schemaspy/SchemaAnalyzer.java
+++ b/src/main/java/org/schemaspy/SchemaAnalyzer.java
@@ -38,7 +38,6 @@ import org.schemaspy.util.ResourceWriter;
 import org.schemaspy.view.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.stereotype.Component;
 
 import java.io.File;
 import java.io.FileFilter;
@@ -56,7 +55,6 @@ import java.util.stream.Collectors;
  * @author John Currier
  * @author Nils Petzaell
  */
-@Component
 public class SchemaAnalyzer {
     private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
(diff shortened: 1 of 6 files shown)


rafalkasa:
@npetzall I would like to understand why Spring class with @Configuration is better than Spring stereotypes ? One of advantage is that Spring will not have to do Component scans when context start and second is that configuration is in Java class, but what are the other advantages ?

npetzall:
Well the real idea behind it is that we encapsulate the usage of spring to the application and everything that we might be libraries doesnt have a dependency on spring. Since not everybody wants spring on their classpath.

There are ways around the having Spring as a dependency when using annotations, since annotations might not be used you could specify annotations as provided. I think, but I believe they will be needed during compile-time(believe).

So it's more of a structural thing. Having stereotypes all over the place which more or less resorts to autowire hell, instead of a readable config. Aka, config can be read prior to runtime, componentscan will be resolved during runtime. You'll never know what you'll get.

But as I said in the beginning, modules, libraries and #157

rafalkasa:
Could you explain me how someone will be using the SchemaSpy as library but will not using the dependencies in code that Spring is resolving as IoC ? 

Maybe I don't understand what you invent, but I can imagine that some one is using SchemaSpy as library to use core SchemaSpy modules / functions for retrieve as a collection all tables from database with the details.

npetzall:
I'll write a longer explanation later but this comment is pretty on point. https://github.com/schemaspy/schemaspy/issues/157#issuecomment-335926877

rafalkasa:
@npetzall before you will spend your time to write explanation form me please let me read the #157 comments

npetzall:
A sort explanation would be that someone that wishes to use schemaspy or modules from schemaspy would wire them up themselves.

Including spring might cause problems even for people using spring. Since they might be using a different version. (They are pretty compatible, but they would need to dissect our jar to remove the spring code, unless we create multiple jars).

rafalkasa:
@npetzall now I understand a little better the idea behind the changes and advantage that offer `@Configuration` classes. Thank you for explanation. I will merge your changes

PR #236, opened 2018-01-24
Removed dependency on spring from Config

npetzall:
* Added test for loadProperties

### DIFF
--- a/src/main/java/org/schemaspy/Config.java
+++ b/src/main/java/org/schemaspy/Config.java
@@ -31,8 +31,6 @@ import org.schemaspy.view.DefaultSqlFormatter;
 import org.schemaspy.view.SqlFormatter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.util.FileCopyUtils;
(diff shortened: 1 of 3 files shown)


rafalkasa:
@npetzall could you explain me why is better to write own code instead use of existing Spring modules/functions ? What is your main idea to remove from Config.class Spring dependency ?

npetzall:
I'll have to admit that Config.java might not be required to remove the dependency to spring.

But I think there are two camps when it comes to software development. One that is always try to use a library, second one is to use libraries when needed. I felt it unnecessary to have a dependency to spring for two simple things that can be easily done with standard java.

In the greater scheme of things, I wonder if spring is actually needed at all. But in the library sense of things, we should limit the usage to application and not library. I could drop the spring removal thing and keep the tests. Since I'm of the believe that we will replace Config.java with smaller config-files and use them with JCommander.

rafalkasa:
@npetzall thank you for your explanation I will merge your changes. Sometimes we are using the libraries when simple Java code is enough and resolve the problem.

PR #238, opened 2018-01-25
Adding build info and removing buildNumber.properties

npetzall:
* Adding git commit hash instead of a number.
* Added additional information to manifest.
* Added BuildInfo to access manifest attributes.
* Made Revision look in correct jar for manifest.
* Fixed issue with value lookup in Revision getOrDefault expects
  key to be Attributes.Name object.
* Added version-info argument to commandline

Revision now returns 6.0.0-rc2 #46120e0 - 2018-01-25 01:43:01+01
-version-info outputs:
image

### DIFF
--- a/buildNumber.properties
+++ b/buildNumber.properties
@@ -1,3 +1,0 @@
-#maven.buildNumber.plugin properties file
-#Tue Dec 12 15:37:44 CST 2017
-buildNumber=40
--- a/pom.xml
+++ b/pom.xml
@@ -34,11 +34,17 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+
         <sonar.junit.reportPaths>target/surefire-reports,target/failsafe-reports</sonar.junit.reportPaths>
 
         <spring-boot.version>1.5.9.RELEASE</spring-boot.version>
 
         <java.version>1.8</java.version>
+
+        <maven.buildNumber.timestampFormat>{0,date,yyyy-MM-dd HH:mm:ssX}</maven.buildNumber.timestampFormat>
+        <maven.buildNumber.useLastCommittedRevision>false</maven.buildNumber.useLastCommittedRevision>
+        <maven.buildNumber.shortRevisionLength>7</maven.buildNumber.shortRevisionLength>
+        <maven.buildNumber.revisionOnScmFailure>unknown</maven.buildNumber.revisionOnScmFailure>
     </properties>
     <issueManagement>
         <url>https://github.com/schemaspy/schemaspy/issues</url>
@@ -243,6 +249,15 @@
                 <filtering>true</filtering>
             </resource>
         </resources>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>buildnumber-maven-plugin</artifactId>
+                    <version>1.4</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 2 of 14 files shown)


rafalkasa:
@npetzall I have problem after when I try to run code from your branch:

```
.ERROR - C:\develop\schemaspy\target\classes (Access is denied)
java.io.FileNotFoundException: C:\develop\schemaspy\target\classes (Access is denied)
	at java.util.zip.ZipFile.open(Native Method)
	at java.util.zip.ZipFile.<init>(ZipFile.java:225)
	at java.util.zip.ZipFile.<init>(ZipFile.java:155)
	at java.util.jar.JarFile.<init>(JarFile.java:166)
	at java.util.jar.JarFile.<init>(JarFile.java:103)
	at org.schemaspy.util.JarFileFinder.findJarFileForClass(JarFileFinder.java:17)
	at org.schemaspy.Revision.initialize(Revision.java:47)
	at org.schemaspy.Revision.<clinit>(Revision.java:43)
	at org.schemaspy.view.DotFormatter.writeHeader(DotFormatter.java:232)
	at org.schemaspy.view.DotFormatter.writeRelationships(DotFormatter.java:291)
	at org.schemaspy.view.DotFormatter.writeRealRelationships(DotFormatter.java:264)
	at org.schemaspy.SchemaAnalyzer.generateHtmlDoc(SchemaAnalyzer.java:309)
	at org.schemaspy.SchemaAnalyzer.analyze(SchemaAnalyzer.java:218)
	at org.schemaspy.SchemaAnalyzer.analyze(SchemaAnalyzer.java:98)
	at org.schemaspy.Main.runAnalyzer(Main.java:93)
	at org.schemaspy.Main.run(Main.java:86)
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:732)
	at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:716)
	at org.springframework.boot.SpringApplication.afterRefresh(SpringApplication.java:703)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:304)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1118)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1107)
	at org.schemaspy.Main.main(Main.java:59)
```

npetzall:
Switched logging from error to warn, it should use a fallback that is Unknown.
Problem should only be visible when running in a non-jar fashion (from IDE)

rafalkasa:
@npetzall after the your update I have the same issue as before. When I'm running SchemaSpy from IDE exceptions occurs. Everything looks good when I'm running from command line. Do you know how to fixed ?

I don't want merge the code that is not working because before everything works without any problems....

npetzall:
Does execution fail or is it only the logging.

What can be done is to check if we are running in jar or not.

But since running in non-jar mode isn't standar I would feel it's unnecessary code, not actually supported.

rafalkasa:
@npetzall I will not change my mind, for me is not difference if code is running from jar or from IDE if exception occur for me is a signal that maybe we choose not proper implementation and is better to rethink and reinvent the solution...

npetzall:
Well there is a significant difference, but I digress we have lots of jar VS non-jar code where non-jar code only runs in development. The best thing I could do is hide the exception in non-jar.

Since running from IDE like you're doing isn't running a proper build. You could do it by running an actual build and you wouldn't get an exception. Manifest is only created when maven package goal is running.

If we hade my other pr you wouldn't see the stacktrace/exception unless you enabled debug.

rafalkasa:
so please add this check when application is run as non-jar code and this is acceptable, second option is to merge first the PR #249 and the problem should also disappear because exceptions will be displayed only when application will be run with `-debug` param.

rafalkasa:
@npetzall I merged PR #249 so could you merge your branch `npetzall:build_info` with master ? I will check how everything is working when I run without or with `-debug `option.

npetzall:
I'll rebase as soon as I get a chance. Alone with both kids.

rafalkasa:
image

npetzall:
I tried to find an image with a dad and two screaming children. Couldn't find it.

npetzall:
Added handling for non-jar executions which looks like this with -version-info
image

npetzall:
Fixing, rerunning tests locally, push soon...

npetzall:
Closing this, will do this later when split between CLI and Library has been done.

PR #249, opened 2018-02-11
Only log stacktraces when -debug or --debug

npetzall:
I read a comment that we might not want to print stacktraces in the faces of our users.
So just for fun I added a logback ThrowableConverter that only prints stacktrace if org.schemaspy logger is in level debug (which it is either by -debug or --debug)
Also altered some logging to make a good use-case.

So with no debug:
image

And with -debug:
image

The rational would be that we could add stacktraces to all logging and just have them skipped in normal flow and have them added in debug.

This would be to avoid
```
LOGGER.warn("Something went wrong");
LOGGER.debug("this is why", exception);
```

### DIFF
--- a/src/main/java/org/schemaspy/Config.java
+++ b/src/main/java/org/schemaspy/Config.java
@@ -929,7 +929,9 @@ public final class Config {
             try {
                 tableInclusions = Pattern.compile(strInclusions);
             } catch (PatternSyntaxException badPattern) {
-                throw new InvalidConfigurationException(badPattern).setParamName("-i");
+                throw new InvalidConfigurationException(badPattern)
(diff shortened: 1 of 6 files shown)


npetzall:
If this looks ok, and is wanted let me know and I'll add a test and also refactor some logging else just close.

rafalkasa:
@npetzall your change looks pretty good. Please add missing tests and we can merge the code.

Ihave only one request could you add some documentation for the SchemaSpy users how they should debug the application if something will go wrong.

Thanks in advance

npetzall:
What do you say about, adding instruction for debugging in the docs but also in the big catch clause thingy in Main.java?

PR #262, opened 2018-02-21
Changes to resolve https://github.com/schemaspy/schemaspy/issues/233

### LINKED ISSUE #233: DatabaseName changes between pages
## Expected Behavior
DatabaseName should be consistent.

## Current Behavior
Index/Table:  
image
Column:  
image

## Possible Solution
Use same single method to getDatabaseName possible in super class.

uppalabharath:
Hi @npetzall / @rafalkasa , 

I have made the changes to resolves #233 . Please review and let me know if any further modifications are needed. 

Thanks in advance!!


src/main/java/org/schemaspy/view/HtmlMainIndexPage.java
```diff
@@ -98,7 +102,7 @@ public int compare(Table table1, Table table2) {
 
         scopes.put("tables", mustacheTables);
         scopes.put("database", database);
-        scopes.put("databaseName", databaseName);
+        scopes.put("databaseName", getDatabaseName(database));
```
npetzall: Shouldn't this line be removed and the getDatabaseName(database) be used in construction of MustasheWriter

uppalabharath: 1. All the usages of mustachewriter.write send the databasename directly by getting the value from getDatabaseName() from HtmlFormatter
2. If the getDatabaseName() method has to be used here in MustacheWriter either the code need to be duplicated from HtmlFormatter or MustacheWriter have to extend HtmlFormatter.

uppalabharath: Yes that can be done. 
Let me make the changes and try.

npetzall:
If it wouldn't be to much ask and you're making changes in HtmlFormatter.
Could you have a look at the issues reported by SonarQube?
https://sonarcloud.io/code?id=org.schemaspy%3Aschemaspy&selected=org.schemaspy%3Aschemaspy%3Asrc%2Fmain%2Fjava%2Forg%2Fschemaspy%2Fview%2FHtmlFormatter.java

uppalabharath:
@npetzall , Yep I shall have a look and get in touch with you for any info I need.

src/main/java/org/schemaspy/view/HtmlMainIndexPage.java
uppalabharath: Hi @npetzall, 

We can pass the getDatabaseName(databse) in the MustacheWriter construction  in HtmlMainIndexPage.java but cannot remove databaseName from scope. 

This variable is used in layout of main.html and is flushed to `result` before the mainScope is added with databaseName. 

```
            Mustache mustache = mf.compile(getReader(templatePath),"template");
            mustache.execute(result, scopes).flush();
```

npetzall: Yepp, saw it now. Feels sort of awkward but it works so.

npetzall: Since if mainScope was created first it would only add unused variables to the main rendering or am I still wrong about everything :smile:, frontend is not my cup of 🍵

uppalabharath:
Hi @rafalkasa / @npetzall , 

Regarding the sonar cube comments, the tool shows three issues. 

1. Removing isMetered field -- No usage can be removed
2. Replace getPathtoRoot method with a constant -- Can be ignored as this method would be used in child classes like HtmlRoutines.java whether the rrot path differs from the default. 
3. Test case coverage for sourceForgeLogoEnabled() method -- No usage found across the project. Can be removed
4. Replace String concatenation with String.format -- can be done

Please review and let me know if you agree with my comments. 

P.S: Is there a way that I can do the Sonar Qube analysis on the github itself. 

Thanks 
Bharath.

npetzall:
This is now handled in #344

PR #264, opened 2018-02-22
Fixed reading from springboot jar.

npetzall:
ERROR - null
java.nio.file.FileSystemNotFoundException: null
	at com.sun.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:171)
	at com.sun.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:157)
	at java.nio.file.Paths.get(Paths.java:143)
	at org.schemaspy.db.config.PropertiesFinder.findClassPath(PropertiesFinder.java:61)
	at org.schemaspy.db.config.PropertiesFinder.find(PropertiesFinder.java:33)
	at org.schemaspy.db.config.PropertiesResolver.getDbProperties(PropertiesResolver.java:41)
	at org.schemaspy.Config.getDbProperties(Config.java:1348)
	at org.schemaspy.service.SqlService.connect(SqlService.java:46)
	at org.schemaspy.SchemaAnalyzer.analyze(SchemaAnalyzer.java:178)
	at org.schemaspy.SchemaAnalyzer.analyze(SchemaAnalyzer.java:97)
	at org.schemaspy.Main.runAnalyzer(Main.java:80)
	at org.schemaspy.Main.run(Main.java:73)
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:732)
	at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:716)
	at org.springframework.boot.SpringApplication.afterRefresh(SpringApplication.java:703)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:304)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1118)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1107)
	at org.schemaspy.Main.main(Main.java:56)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:48)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:87)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:50)
[...]

PR #266, opened 2018-02-25
Use built-jar and fix error

npetzall:
First commit makes failsafe (integration tests use built jar)

This will faile.

After the failure, I'll add the fix that exists in PR #264


npetzall:
So now I has failed. By just changing classpath of failesafe-plugin.
Adding fix for reading from jar (FileSystem Zip needs to be created it will not be created on the fly, however it might already exist)

PR #283, opened 2018-03-17
Removed unused ProgressListener from Database model

npetzall:
Progress should be around the model not in the model.


rafalkasa:
Of course, looks good. Thank you for your contribution and make the SchemaSpy better and better

PR #284, opened 2018-03-18
Added nohtml to CommandLineArguments

npetzall:
* nohtml is not an output attribute but an application/processing

End goal is to move progresslistener from the creation in SchemaAnalyzer to a bean, so it can be injected in multiple places and not passed as argument.

Later on this will help when switching from current progresslistener to an implementation with commandline progressbar.


rafalkasa:
For me changes are ok

Issue #267, opened 2018-02-25
Make getExportedKeys optional


npetzall:
## Expected Behavior
Everything should finish quickly.

## Current Behavior
With 1000 tables it's not quick at all.

## Possible Solution
for mysql (this is probably implementation specific)
The getExportedKeys in TableService.connectTables is super slow.
Comment at the beginning of getExportedKeys
```
// also try to find all of the 'remote' tables in other schemas that
// point to our primary keys (not necessary in the normal case
// as we infer this from the opposite direction)
```

So it might be desirable to have it optional, and that you would need to enable it.

Silly me, three rows down the following comment exists
```
// get the foreign keys that reference our primary keys
// note that this can take an insane amount of time on Oracle (i.e. 30 secs per call)
```
Which probably explains #250 
3500 tables * 30 seconds per table estimated time would be about 29 hrs to complete.

## Steps to Reproduce (for bugs)

1. Create lots of tables Oracle or Mysql (both are oracle haha)
2. Try to analyze
3. Wait
4. Wait
5. Wait

## Context

* Database type (e.g. MySQL): MySQL
* Database driver version (is you use a special one): mysql-connector-j 6.0.6
* Your DDL (sample how to reproduce the issue): Just lots of tables, they don't even need to be connected.
* Full command including arguments how you run SchemaSpy: standard commandline for mysql, no frizzles

## Your Environment

* Version used: snapshot
* Environment name and version (e.g. Chrome 39): 
* Operating System and version (e.g. Windows 10): Ubuntu 16.04
* Java version: 1.8

npetzall:
Comment in code:
```
// also try to find all of the 'remote' tables in other schemas that
// point to our primary keys (not necessary in the normal case
// as we infer this from the opposite direction)
// get the foreign keys that reference our primary keys
// note that this can take an insane amount of time on Oracle (i.e. 30 secs per call)
```
From JavaDoc DatabaseMetaData.getExportedKeys()
```
* Retrieves a description of the foreign key columns that reference the
* given table's primary key columns (the foreign keys exported by a
* table
```

getExportedKeys() get information about tables that have a FK pointing towards the specified table.
(this information is specified in another table than the requested)

getImportedKeys() get information about tables that the specified table has a FK towards.
(this information is specified in the table, that you are requesting)

So by default getImportedKeys should be much quicker. But I also assume that this information might be easier to extract with a custom query.

[...]

PR #290, opened 2018-03-23
Added -noexportedkeys option to avoid lengthy calls to getExportedKeys()

npetzall:
* Added clock to DatabaseService to test time based code.
* If Estimated time for completion of connectTables.views/tables is
  longer than 30 minutes a logging will appear that the option
  -noexportkeys exists and the implications of using it.

Solves issue #267


npetzall:
Hmm, works locally retrigge

PR #302, opened 2018-04-10
Never use both SQL and DatabaseMetaData in DatabaseService.getBasicTableMeta

### LINKED ISSUE #222: -all on mysql adds tables from one schema to another in the output
## Expected Behavior
Tables from schema/database CME should not end up in information_schema.

## Current Behavior
Table cmeApiQuerySetting is added as a table in schema 'information_schema'.

## Possible Solution
The org.schemaspy.service.DatabaseService#getBasicTableMeta first tries with custom query.
Since information_schema doesn't have tables according to custom query.
DatabaseMetaData.getTables() is called. Since Mysql has done their own take on schema/database/catalog.
(yes, they call it schema/database in documentation for mysql and in jdbc it seems to be catalog/database)
So it seems that information_schema might include all datatabases/schemas since it searches catalog = null and mysql jdbc totally ignores schema.

So either only go with DatabaseMetaData.getTables if missing SQLquery or the query fails(SQLException)

Or implement some kind of strategy(remapping) that can adapt Catalog/Schema using information in databaseType. So they both end up with 0 tables.

SQL Script
```
CREATE DATABASE CME;

CREATE TABLE CME.cmeApiQuerySetting (
  cmeApiQSid INTEGER AUTO_INCREMENT,
  name varchar(40) NOT NULL,
  value varchar(40) NOT NULL,
  PRIMARY KEY (cmeApiQSid),
  UNIQUE key_unique (name)
) engine=InnoDB;

GRANT SELECT on `CME`.* to test@`%`;
FLUSH PRIVILEGES;
```
Commandline args
```
docker run -v "$PWD:/output" schemaspy/schemaspy:snapshot -t mysql -u root -p test -db CME -all -host [host] -port [port]
```

## Context
* Database type (e.g. MySQL): mysql:5.6 docker image from docker hub.
* Database driver version (is you use a special one): mysql-connector-java-6.0.6.jar
```
CREATE DATABASE CME;

CREATE TABLE CME.cmeApiQuerySetting (
  cmeApiQSid INTEGER AUTO_INCREMENT,
  name varchar(40) NOT NULL,
  value varchar(40) NOT NULL,
  PRIMARY KEY (cmeApiQSid),
  UNIQUE key_unique (name)
) engine=InnoDB;

GRANT SELECT on `CME`.* to test@`%`;
FLUSH PRIVILEGES;
```
```
docker run -v "$PWD:/output" schemaspy/schemaspy:snapshot -t mysql -u root -p test -db CME -all -host [host] -port [port]
```

## Your Environment

* Version used: docker schemaspy/schemaspy:snapshot (2018-01-16)
* Environment name and version (e.g. Chrome 39): N/A
* Operating System and version (e.g. Windows 10): Docker
* Java version: Docker

npetzall:
So, if SQL exists use it.
if SQL fails use DatabaseMetaData.
if SQL doesn't exist use DatabaseMetaData.

Problem isolated to mysql since they ignore SchemaPattern
(yes, they call it catalog)

Closes #222 with option 2

### DIFF
--- a/src/main/java/org/schemaspy/service/DatabaseService.java
+++ b/src/main/java/org/schemaspy/service/DatabaseService.java
@@ -57,7 +57,7 @@ public class DatabaseService {
 
     private final SqlService sqlService;
 
-    private final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+    private static final Logger LOGGER = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
     public DatabaseService(TableService tableService, ViewService viewService, SqlService sqlService) {
         this.tableService = Objects.requireNonNull(tableService);
@@ -95,7 +95,7 @@ public class DatabaseService {
         updateFromXmlMetadata(config, db, db.getSchemaMeta());
     }
     
-   private void initCatalogs(Database db) throws SQLException {
+    private void initCatalogs(Database db) throws SQLException {
 
             String sql = Config.getInstance().getDbProperties().getProperty("selectCatalogsSql");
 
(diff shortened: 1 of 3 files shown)


npetzall:
When this might happen for other dbms, the schemaPattern will be affective and it won't find anything. Since mysql ignores//has mixed catalog and schema in JDBC driver, it wont narrow the search to a schema. It will find everything.

PR #303, opened 2018-04-12
Replaced PropertyFileDefatulProvider with a CombinedDefaultProvider

npetzall:
* Combined will have search order that ends with EnvDefaultProvider
* If property file is available it will be used first
* No property file only env will be used


npetzall:
We should decide if we continue with JCommander or switch to picocli.
It will include some bigger changes so it will end up in 7.0.0

PR #308, opened 2018-04-17
Added test for HtmlEncoder

npetzall:
* Added dependency for JunitParams


npetzall:
Started to look at SqlFormatter and it isn't used will be removed in other PR.
When removed HtmlEncoder is no longer used and will be removed. So no need to test stuff that isn't used and will be removed.

PR #314, opened 2018-04-21
Big standard outputs to logger

npetzall:
Replaced a lot of System.out/err with Logger
Some other minor fixes that were quick to do
Also added sonarqube ignores to things we won't fix in place.

Added #313 so that I could run tests

This is needed to for ProgressBar development.


npetzall:
Closing this, we should handle it as the come along or directly when progress bar will be fully implemented.

Issue #318, opened 2018-04-23
Improve integrationtesting time


npetzall:
It should be possible to cut the time for integrationtesting by only using a single mysql container and multiple schemas. So one container and schema per test.

This could also be base for a multi-schema test.

Since only one container is used per dbms, it should cut down the testing time.
We could also look at running tests in parallel.
But that might need refactoring of production code since there are lots of static instances.

PR #319, opened 2018-04-25
318 integration test speed

npetzall:
By running as a suite and having the JdbcContainerRule in the suite class we get one container for all mysql tests.

Non-scientific testing reveals that the testing time should be reduced by 40-50 seconds.


npetzall:
More speed improvements might be possible with JUnitCore.runClasses(Computer, Class<?>...classes) but I tried it and it actually performed worse. Might have to do with synchronization. There are lots of static classes (Config) that is modified between tests. I would assume that we could give it a go, when we have moved away from mutable static instances.

npetzall:
Relates to #318

PR #321, opened 2018-04-28
Simplified the Hash sent to Mustache in HtmlConstraintPage

### LINKED ISSUE #320: Check Constraints not showing for MS SQL Server
## Expected Behavior
Table level check constraints in the MS SQL Server database should show up in the output pages.

## Current Behavior
No check constraints are listed.  Foreign Key constraints are listed.

## Possible Solution
With only changing the SchemaSpy Jar... schemaSpy_5.0.0.jar works in showing check constraints ... schemaspy-6.0.0-rc2 does not.  Of course I would like to use version 6x.

1. With only changing the SchemaSpy Jar... schemaSpy_5.0.0.jar works in showing check constraints ... schemaspy-6.0.0-rc2 does not.

2.  I've tried various *.properties  for mssql you have posted.   both with the 5x schemaspy and with 6x. No change in behavior (accept when it can run the older mssql type vs. the newer mssql05)

## Context

I really want to use this for our QA environment to help us better work with our databases but without listing the check constraints I can use it.

* Database type (e.g. MySQL): mssql05-jtds
* Database driver version (is you use a special one): jtds-1.3.1.jar

-jar "C:\Users\kfugate\Downloads\SchemaSpy\schemaSpy_5.0.0.jar" -gv "C:\Program Files (x86)\Graphviz2.38" -dp "C:\Users\kfugate\Downloads
\SchemaSpy\jtds-1.3.1-dist\jtds-1.3.1.jar" -t mssql05-jtds -db Customer -s dbo -host MYHOST -port 1433 -u myusername -p mypassword -connprops "C:\Users\kfugate\Downloads
\SchemaSpy\properties\mssql-jtds.properties" -I "syncobj.*" -o "C:\TMP\SchemaSpy\18_03\customer" -lq

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39):
* Operating System and version (e.g. Windows 10):  Windows Server 2012r2
* Java version: 
JAVA_VERSION="1.8.0_161"
OS_NAME="Windows"
OS_VERSION="5.1"
OS_ARCH="i586"

npetzall:
* map.entrySet isn't available unless you change the ReflectionHandler
  so preparing the data is actually better since Inverted sections
  becomes easier.
* Added mssqlserver tests (trying to prepare for a suite)
* Put some duplicated Validation code into a separate classes

* fixes #320


npetzall:
I'm unsure about the comment 
> I understand the idea behind, but we are sending the same content few times to for test producesSameContent. Do you have any other options or solution for that ?

npetzall:
The end-to-end tests are really bad.
I've looked into how we can avoid them there is a new project that is for the purpose to remove them.
The structure or at least that "fields" are populated in HTML is true by just using a fully populated database model.

So the HTML/XML output should only tested with a database model irregardless of dbms. So that the binding between model and html/xml works, this should also be done on a page by page level and not as a whole. This could be tested by just verifying that output contains the data that is in the model.

As for dbms tests, they should only populate a model and the model should be verified.

So changes that needs to be done.
* No code should be able to access dbms thru database model (remove DatabaseMetaData from model)
* HTML/DOT/XML should be done using a supplied Stream/Writer and not a target file (don't want to hit the file system just to check what output would be generated, ByteArrayOutputStream/StringWriter or FileOutputStream/FileWriter.
* Remove static calls and use constructor arguments so that dependencies can easily be mocked (want to test in isolation to reduce the permutation of arguments)

Issue #305, opened 2018-04-15
Changing License to LGPLv3 or later for the project


npetzall:
Backstory can be read in issue #206 

Original project was under LGPL 2.1 as written on the webpage.
Source contains following text
```
 * SchemaSpy is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation; either
 * version 2.1 of the License, or (at your option) any later version.
```

Which makes it version 2.1 or later.

Due to incorrect licensing see issue #206 your contributions have probably implicitly been licensed under MIT.

According to https://www.dwheeler.com/essays/floss-license-slide.html

MIT and LGPLv2.1 or later can be combined under "LGPLv2.1 or later" or "LGPLv3 or later"

If you look again at https://www.dwheeler.com/essays/floss-license-slide.html. 
You also see that Apache 2.0 is not compatible with LGPLv2.1 or later.

The project has a dependency to at least one Apache 2.0 licensed library. 
So the project as a whole must be licensed under "LGPLv3" or "LGPLv3 or later.".

It would be more clear if all source code in the project was under the same license.

We ask that you grant the SchemaSpy project your contributions under LGPLv3 or later.
(we will add copyright and license information to source code-files)

You can grant us your contributions under LGPLv3 or later by commenting to this issue with the following text (replace "[name]" with your full legal name).

```
I, [name] hereby grant the SchemaSpy project all of my contributions, past and present, to the SchemaSpy project
under GNU Lesser General Public License as published by the Free Software Foundation; either
version 3 of the License, or any later version.
```

**Contributors:**

- [x] @rafalkasa
- [x] @npetzall
- [x] @danwatt
- [x] @Riggs333
- [x] @wkasa 
- [x] @ismailsimsek
- [x] @mapausch
- [ ] @ClintEstwood
- [x] @martenbohlin
- [x] @johanvdw
- [x] @jornh
- [x] @dhermanns
- [x] @uppalabharath
- [x] @alpha0010
- [x] @jholland-usgs
- [ ] @igor-suhorukov

If you don't approve of this change we will add MIT Notice to files that have be contributed by you.
And we will distribute it under LGPLv3 or later.

johanvdw:
I, Johan Van de Wauw hereby grant the SchemaSpy project all of my contributions, past and present, to the SchemaSpy project under GNU Lesser General Public License as published by the Free Software Foundation; either version 3 of the License, or any later version.
--
[...]

PR #322, opened 2018-04-29
Adding copyright and license notice

npetzall:
Adding copyright and license notice to src-files
Removed license-file, using LGPL COPYING and COPYING.LESSER
I haven added any license info to src/**/resources except for banner.txt

We might want to add it to templates but I'm not sure.

This is a cover-all so there might be copyright notice when in fact the changes might be to small (or might later have resulted only in  a empty line) to deem copyrightable.

I haven't changed the old license notice from LGPLv2.1 or later to LGPLv3 or later. I think it's possible but there is no actual need for it.

@rafalkasa I haven't done any diff against the sourceforge repo, so If you did any changes or extracted any code into new classes before the initial commit to this repo. We need to add John Currier as a copyright holder in affected files. Same goes for files that might be missing your copyright notice if they were changed before initial commit.

I didn't add copyright and author if only changes where package name.


npetzall:
Related to issue #305

npetzall:
@rafalkasa 
Just need OK from you on two things that affect end-user
1, Banner has changed to:
```
  ____       _                          ____
 / ___|  ___| |__   ___ _ __ ___   __ _/ ___| _ __  _   _
 \___ \ / __| '_ \ / _ \ '_ ` _ \ / _` \___ \| '_ \| | | |
  ___) | (__| | | |  __/ | | | | | (_| |___) | |_) | |_| |
 |____/ \___|_| |_|\___|_| |_| |_|\__,_|____/| .__/ \__, |
                                             |_|    |___/

                                              

SchemaSpy generates an HTML representation of a database schema's relationships.
SchemaSpy comes with ABSOLUTELY NO WARRANTY.
SchemaSpy is free software and can be redistributed under the conditions of LGPL version 3 or later.
http://www.gnu.org/licenses/
```
Additional 3 lines regarding warrenty, license and the link to licenses.

2, New commandline option `-l` or `--license` which prints GPL and LGPL in the terminal window.

If they are OK with you I'll merge right away.

npetzall:
Been a week merging.

PR #323, opened 2018-05-08
Update mssql05.properties to include schema to prevent empty table comments

jwstadler:
In the default mssql05.properties file, the selectTableCommentsSql query does not include the schema when building the Object ID which causes joins to fail and comments to be skipped for tables. Adding the schema name resolves the issue.

### DIFF
--- a/src/main/resources/org/schemaspy/types/mssql05.properties
+++ b/src/main/resources/org/schemaspy/types/mssql05.properties
@@ -24,7 +24,7 @@ dbThreads=1
 
 # return table_name, comments for current schema
 # SQL provided by Stefano Santoro
-selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0
+selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0 AND i_s.table_schema \= :schema
 
 # reference: http://databases.aspfaq.com/schema-tutorials/schema-how-do-i-show-the-description-property-of-a-column.html
 # return table_name, column_name, comments for a specific :schema
--- a/src/main/resources/org/schemaspy/types/mssql05.properties
+++ b/src/main/resources/org/schemaspy/types/mssql05.properties
@@ -24,7 +24,7 @@ dbThreads=1
 
 # return table_name, comments for current schema
 # SQL provided by Stefano Santoro
-selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '..' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0
+selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0
 
 # reference: http://databases.aspfaq.com/schema-tutorials/schema-how-do-i-show-the-description-property-of-a-column.html
 # return table_name, column_name, comments for a specific :schema
--- a/src/test/java/org/schemaspy/testcontainer/MSSQLServerIT.java
+++ b/src/test/java/org/schemaspy/testcontainer/MSSQLServerIT.java
@@ -105,6 +105,12 @@ public class MSSQLServerIT {
     }
 
     @Test
+    public void databaseShouldBePopulatedWithTableTestWithComment() {
+        Table table = getTable("TestTable");
+        assertThat(table.getComments()).isEqualToIgnoringCase("This is a table description");
+    }
+
+    @Test
     public void databaseShouldBePopulatedWithTableTestAndHaveColumnName() {
         Table table = getTable("TestTable");
         TableColumn column = table.getColumn("Description");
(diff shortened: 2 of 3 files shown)


npetzall:
Cool, could you possibly improve on the mssqlserver test we have. Just checked and it only covers column comments.

Would be nice if it covered catalog, schema and table. 

If you can't, I'll create an issue so that we won't forget.

jwstadler:
Hi Nils,

 

Let me take a look at it this evening.

npetzall:
If you need any help or have questions I'm regularly available on gitter, there's a badge in the overview.

jwstadler:
I added a test to cover table comments for MSSQL and updated the MSSQL db script to add a test comment for the table.

npetzall:
I copied the test to master and it works without the fix.
The only way that it could be invalid is with multipl schemas where catalog and table is same.

`OBJECT_ID` function will only return first match.

So
`OBJECT_ID(i_s.table_catalog + '..' + i_s.table_name)` 
Will return same object id for all schemas (the first it finds)

So with the change
`OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name)`
It will return comments from each schema. Sadly since this isn't filtered on schema.
You might end up with the wrong comment, since it will select the last.

The below will work without your change, and will fail with the change.

Test method:
```
@Test
    public void databaseShouldBePopulatedWithTableTestWithComment() {
        Table table = getTable("TestTable");
        assertThat(table.getComments()).isEqualTo("This is a test table");
    }
```
Test script:
```
CREATE DATABASE TEST;

USE TEST;

CREATE SCHEMA test1;

CREATE TABLE TestTable
(
 Id INT PRIMARY KEY IDENTITY(1,1) NOT NULL,
 Name NVARCHAR(40) NOT NULL,
 Description NVARCHAR(100) NOT NULL,
);

EXEC sys.sp_addextendedproperty
    @name=N'MS_Description',
    @value=N'This is a test table' ,
    @level0type=N'SCHEMA',@level0name=N'dbo',
    @level1type=N'TABLE',@level1name=N'TestTable'

EXEC sys.sp_addextendedproperty
    @name=N'MS_Description',
    @value=N'This is column description' ,
    @level0type=N'SCHEMA',@level0name=N'dbo',
    @level1type=N'TABLE',@level1name=N'TestTable',
    @level2type=N'COLUMN',@level2name=N'Description'

CREATE TABLE test1.TestTable
(
 Id INT PRIMARY KEY IDENTITY(1,1) NOT NULL,
 Name NVARCHAR(40) NOT NULL,
 Description NVARCHAR(100) NOT NULL,
);

EXEC sys.sp_addextendedproperty
    @name=N'MS_Description',
    @value=N'This is a test table in test1' ,
    @level0type=N'SCHEMA',@level0name=N'test1',
    @level1type=N'TABLE',@level1name=N'TestTable'

EXEC sys.sp_addextendedproperty
    @name=N'MS_Description',
    @value=N'This is column description in test1' ,
    @level0type=N'SCHEMA',@level0name=N'test1',
    @level1type=N'TABLE',@level1name=N'TestTable',
    @level2type=N'COLUMN',@level2name=N'Description'
```

npetzall:
Changing:
`selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0`

by appending `AND i_s.table_schema = :schema` will fix the issue together with your fix.

So it should be:
`selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0 AND i_s.table_schema = :schema`

npetzall:
Same problem exists for Column comments.

npetzall:
I wonder how much I can break this. Since it doesn't account for catalog.

jwstadler:
The reason comments were not found for tables in schemas in the original code was due to using the double dot syntax when referencing the object:

```
OBJECT_ID(i_s.table_catalog + '..' + i_s.table_name)
```

Using this syntax only returns Object IDs for tables in the default schema, which is typical `dbo`, unless the default schema is changed. This means that it will not return any Object IDs for tables outside of the default schema. Since the existing test script only uses the `dbo` schema it passes with and without the recommended change. 

The query below can be used to illustrate the difference.

```sql
SELECT
    i_s.TABLE_CATALOG,
    i_s.TABLE_SCHEMA,
    i_s.TABLE_NAME,
    OBJECT_ID(i_s.table_catalog + '..' + i_s.table_name) AS ObjectIdDefaultSchema,
    OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) AS ObjectIdWithSchema
FROM INFORMATION_SCHEMA.Tables i_s 
```

If you execute this query on a MSSQL database with tables in the default schema and named schemas as in the updated test script (with `dbo` and `test1`), you will see that null is returned when the double dot syntax is used and the table is not in the default schema. Using the fully qualified reference, along with the join to the `INFORMATION_SCHEMA.Tables` table ensures the correct Object ID is returned for each table. 

The addition of `AND i_s.table_schema = :schema` to the query seems like a good idea. It should prevent the issue you mentioned above where the same table appears in multiple schemas.

Regarding Issue #324, the column comments appear to be working for me, but I see where the above mentioned `:schema` filter could offer a safety net. I don't believe the catalog will come into play in the MSSQL scenario because the `INFORMATION_SCHEMA.Tables` is limited to the current catalog (database) anyway.

I appreciate you taking the time and reviewing this pull request. SchemaSpy is a great tool.

jwstadler:
It seems like the following query work for the MSSQL column comments:

```sql
SELECT 
    i_c.TABLE_NAME, 
    i_c.COLUMN_NAME,
    CAST(s.value AS VARCHAR(MAX)) AS comments
FROM INFORMATION_SCHEMA.Columns i_c 
INNER JOIN sys.extended_properties s 
    ON s.major_id = OBJECT_ID(i_c.table_catalog + '.' + i_c.table_schema + '.' + i_c.table_name) 
    AND s.minor_id = i_c.ORDINAL_POSITION
    AND s.name = 'MS_Description' 
WHERE i_c.TABLE_SCHEMA = :schema
```

npetzall:
I probably jumped the gun on how object_id worked. I assumed since dbo was first result, was a bit colored by schemaspy code that uses the last result.

src/main/resources/org/schemaspy/types/mssql05.properties
```diff
@@ -24,7 +24,7 @@ dbThreads=1
 
 # return table_name, comments for current schema
 # SQL provided by Stefano Santoro
-selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '..' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0
+selectTableCommentsSql=SELECT i_s.TABLE_NAME, CAST(s.value AS VARCHAR(MAX)) AS comments FROM INFORMATION_SCHEMA.Tables i_s INNER JOIN sys.extended_properties s ON s.major_id \= OBJECT_ID(i_s.table_catalog + '.' + i_s.table_schema + '.' + i_s.table_name) WHERE s.class \= 1 AND s.name \= 'MS_Description' AND s.minor_id \= 0
```
npetzall: need to append `AND i_s.table_schema = :schema` to end of query to make table comments test in #325 to pass.

Issue #324, opened 2018-05-09
Column comments query for mssql aren't schema and catalog aware


npetzall:
## Expected Behavior
I expect that a catalog.schema.table.column should have the correct comment 

## Current Behavior
Due to the current SQL it doesn't filter on catalog or schema so it's an ordering issue.
Only if the same comment exists in another catalog or schema.

## Possible Solution
Add catalog and schema to where clause

PR #325, opened 2018-05-09
Added tests for comments in sqlserver

npetzall:
This will test comments on catalog, schema, table and columns in a sql server instance with multiple databases(catalogs),  with multiple schemas in same database(catalog) where all tables and columns have the same name.

Since this in done on master without modifications all tests fails.

Happy hunting!


npetzall:
fixes #324

npetzall:
PR 323 should be merge first and then this rebased.

PR #327, opened 2018-05-16
Version 6.0.0-rc2 to 6.0.0-SNAPSHOT so latest builds will be snapshot

### LINKED ISSUE #139: Version strategy
Since we have a snapshot, it should be called snapshot. Maven has a release plugin we should use it.

npetzall:
Will close #139 for now.

After a release or RC build a second commit will be done updating the version yet again.

After RC build it will be put back to -SNAPSHOT.
After release build it will be set to nextVersion-SNAPSHOT.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <artifactId>schemaspy</artifactId>
     <packaging>jar</packaging>
     <name>SchemaSpy</name>
-    <version>6.0.0-rc2</version>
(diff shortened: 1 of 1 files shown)


rafalkasa:
Yout PR is resolving the issue. 
SNAPSHOT version will be created much more frequently than new release. We should remember to change the version back to SNAPSHOT each time after release.

PR #332, opened 2018-05-21
Main.run doesn't call System.exit()

npetzall:
It closes the SpringContext and evals ExitCodeGenerator so that SpringContext can report an ExitCode. 

But since the returned ExitCode isn't used and we don't use any ExitCodeGenerators and we don't call System.exit with the returned value.

We should probably switch to SpringApplication.close and then use the ReturnCode provided to the exitApplication method in the System.exit call.

First commit with test, proves that there is a problem.
Second commit the fix in production code and tests should be come green.

### DIFF
--- a/src/main/java/org/schemaspy/Main.java
+++ b/src/main/java/org/schemaspy/Main.java
@@ -123,7 +123,8 @@ public class Main implements CommandLineRunner {
     }
 
     private void exitApplication(int returnCode) {
-        SpringApplication.exit(context, () -> returnCode);
+        int computedReturnCode = SpringApplication.exit(context, () -> returnCode);
+        System.exit(computedReturnCode);
     }
 
 }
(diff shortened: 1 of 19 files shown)


npetzall:
So actually calling System.exit() is incompatible with failesafe....
http://maven.apache.org/surefire/maven-failsafe-plugin/faq.html#vm-termination

Need to rewrite some code or test....

npetzall:
Extracted CommandLineRunner from Main to SchemaSpyRunner.
Was awful to test when it had CommanLineRunner interface removed it.
Main now
1, Fetches SchemaSpyRunner bean and calls run.
2, Then exits the SpringApplication
3, Calls System.exit

So all tests that used to call Main.main now retrieves the SchemaSpyRunner and calls it's run
There are separate tests for SchemaSpyRunner and what ExitCode is set.
And Main that System.exit is called.

Issue #165, opened 2017-11-01
Image maps dropped from v5 to v6


mdb3624:
One of the best features in my opinion was the ability to click on an image and being sent to the appropriate table page. This allowed me to better explore the schema when I needed to review a specific area of the database without having to go back to the "main" page

danwatt:
I just tested this using the latest code from master, and it still works. Can you look at one of your table.html files, and see if there is a `<map>` element?

Example:

```html
<map id="oneDegreeRelationshipsDiagram" name="oneDegreeRelationshipsDiagram">
--
  | <area shape="rect" id="node1" href="film.html" title="film" alt="" coords="192,167,512,617">
  | <area shape="rect" id="node2" href="language.html" title="language" alt="" coords="5,301,148,419">
  | <area shape="rect" id="node3" href="film_actor.html" title="film_actor" alt="" coords="564,5,701,152">
  | <area shape="rect" id="node4" href="film_category.html" title="film_category" alt="" coords="556,169,709,316">
  | <area shape="rect" id="node5" href="inventory.html" title="inventory" alt="" coords="561,333,705,509">
  | </map>
```

If it is there and isnt working, what browser are you using?

rafalkasa:
@mdb3624 you are the first SchemaSpy user that is reporting such issue. Could you open in your browser Developers Tools and verify the Console logs. Maybe you have some errors or something.

This is basic SchemaSpy feature and should work also for you 😄

mdb3624:
I am working on a windows 10 using both MS Edge as well as Chrome. Both of these are showing the same behavior. When I open the console I see this error:

Uncaught TypeError: Cannot read property 'value' of null
    at Function.fromTextArea (codemirror.js:9351)
    at table.js:72

BTW, I was working with John when he published this tool so, I realize that it has been here since day one.

Thanks for your help.

mdb3624:
I may have found an issue. Some of the tables in the diagram have "dbo.[table]" these images are not working. The ones that do not have the "dbo." seem to work find. I am using the following server
Database Type: Microsoft SQL Server - 13.00.4001

also I am using the mssql05 databasetype.

rafalkasa:
@mdb3624 thank you for additional information.
I will try to reproduce the issue with specific table name.

This quite strange behaviour when I will reproduce it will be fixed 😄

danwatt:
[...]

PR #339, opened 2018-06-01
Make eval of remote table use database information instead of config

npetzall:
1, Adding test to show the problem
2, Adding fix for #165


src/main/java/org/schemaspy/service/TableService.java
```diff
@@ -353,14 +353,13 @@ protected void addForeignKey(Database db, Table table, String fkName, String fkC
             Table parentTable = tables.get(pkTableName);
 
             String parentContainer = pkSchema != null ? pkSchema : pkCatalog != null ? pkCatalog : db.getName();
-            String catalog = commandLineArguments.getCatalog();
-            String baseContainer = config.getSchema() != null ? config.getSchema() : catalog != null ? catalog : db.getName();
+            String childContainer = table.getSchema() != null ? table.getSchema() : table.getCatalog() != null ? table.getCatalog() : db.getName();
```
npetzall: Doesn't Table have getContainer?
Might be good to extract method so that the way it's creates can be identical in logic.

npetzall: Yeah, it's strange.
Table has 
`this.container = schema != null ? schema : catalog != null ? catalog : db.getName()`
RemoteTable has
`this.baseContainer = baseContainer;`
Since RemoteTable is a subclass of Table it also has container from Table.

And baseContainer is set with all kinds of values that are either for the remoteTable or for the childTable.

Also baseContainer can be either Schema or Catalog.

Issues to create:
1, Rename baseContainer in RemoteTable to something in the line of childContainer, same goes for parameters used to set the field.
2, Refactor to make it catalog aware (Catalog|Database).Schema

npetzall:
This actually fixes a really strange issue I got when specifying no schema or cat when running tests.
It's seems that all tables are considered remote and I end up with a state where a resultset has been closed. RemoteTable creating RemoteTable creating RemoteTable ends with a ResultSet that has been closed. This might either be timeout or there is a recusion issue where the connection/resultset is shared and closed.

PR #340, opened 2018-06-03
Added documentation of command-line arguments to read the docs.

npetzall:
* Includes unimplemented/not used. If deprecated we should write so.


npetzall:
#343 will remove all Unimplemented/note used. So when/if #343 is merged. Documentation needs to be updated.

PR #342, opened 2018-06-04
Replaced all usage of LineWriter with PrintWriter.

npetzall:
Replace custom LineWriter with PrintWriter which is more or less the same.
So there is no need for LineWriter.


npetzall:
It's also a shared class which is bad for isolation and encapsulation.

PR #372, opened 2018-06-27
Added null checks, and this resolved the issue of Null Table columns

zonar-jaybose:


rafalkasa:
@zonar-jaybose thank you for your contribution, before your PR will be reviewed and merged please resolve the UnitTest failures after your change.

npetzall:
@zonar-jaybose @rafalkasa the failed test is actuallymy fault.

I added a Check constraint and didn't specify a name, so the constraint name is generated, and something has recently changed in Travis so that the generated name differs. I've made a fix in my last pr to verify. But things got in the way and I didn`t have the time to move the fix to master.

So I'll fix this later today.

zonar-jaybose:
Understood. I was using the latest code for 6.0 (a download of it, not from source), to build diagrams for 2 separate Postgres 9.5 databases.
With one schema, there was no issue at all.

However, with another, there were null pointer exceptions, with just the word `null`, and no description of what was actually null.

I then pulled master, and built the code, and added debug statements, and null checks in the code to find where the nulls were originating. The final code in the PR was what allowed the failing attempt to proceed. It seems that null `TableColumn`s are being added to collections, at some point.

I agree, this only resolves the symptom.

npetzall:
TableService at the end with initPrimaryKeys it does a get and might add null to primary Keys.

Would it be OK to get a create statement for the table? The weird thing is that it finds the primaryKey constraint but it can't find the column. So there is something strange indeed.

PrimaryKey constraint referencing a non-existing column.

npetzall:
If it was the latest download link in Readme there is a `-debug` option that will print stacktrace. Gonna se if this could be printed at the end if stacktrace should have been printed.

zonar-jaybose:
Just so you know, I ran my attempts with `-debug`, with the downloaded code, and with my forked code (both before my change, and after). The messages were never more descriptive.

npetzall:
Depends on what one finds more descriptive. In normal mode you won't get stacktrace, so with debug you get stacktrace and you'll now somewhat where the problem lies. Which was the only thing I promised stacktrace. But do you have any idea why it became null. Is the table with the problem weird?

zonar-jaybose:
Great question. The 2nd DB and its tables _seem_ fine.

npetzall:
https://github.com/schemaspy/schemaspy/blob/f01286630d95eb648cf31534d18f1fa404e82707/src/main/java/org/schemaspy/service/TableService.java#L649-L681

This is the code that adds the null. Specifically 680 seems to be the only place that can add null. All other places where PrimaryKeys are added are with real objects and nullpointer exception would have happened way before your fix location.

So for this to happen, DatabaseMetaData (JDBC-driver) returns information about primarykeys.
The result should be something like
```
TABLE_CAT, String {@code =>} table catalog (may be <code>null</code>)
TABLE_SCHEM, String {@code =>} table schema (may be <code>null</code>)
TABLE_NAME, String {@code =>} table name
COLUMN_NAME, String {@code =>} column name
KEY_SEQ, short {@code =>} sequence number within primary key( a value
     of 1 represents the first column of the primary key, a value of 2 would
     represent the second column within the primary key).
PK_NAME, String {@code =>} primary key name (may be <code>null</code>)
```

It tries to use the returned COLUMN_NAME to retrieve a previously "imported" column.
But it cant find it, so either the COLUMN_NAME is incorrect, the entire PrimaryKey informtion is incorrect or there was an error "importing" all columns from the table.

I would suggest to add null check in TableService.java#setPrimaryColumn.
```
TableColumn primaryKeyColumn = table.getColumn(columnName);
if (Objects.nonNull(tableColumn) {
  table.setPrimaryKeyColumn(primaryKeyColumn);
} else {
  LOGGER.error("PrimaryKeyColumn '{}' reported in '{}' was not found", columnName, pkName);
}
```
And if someone was to find themselves adventurous I would rename the Table#setPrimaryColumn to addPrimaryColumn.

npetzall:
Closing since it has been replaced by #381 
@zonar-jaybose You should be able to run with snapshot now.
If you still have access to the database that caused the nullpointer issue. It  would be awesome to test the snapshot and send us the error message that I've added.

zonar-jaybose:
@npetzall thank you. Once I get a chance, I will git it a try.

PR #375, opened 2018-07-04
fix: show column details for ophans tables

RMsiemens:
The generated diagrams (png, svg) for orphan tables do not show the details column as the normal diagrams do.
This is actually an important information. And as we want to use schemaspy to generate the pngs for our documentation this information is required.


npetzall:
Won't this just duplicate the diagram already written for the table? In that case we can just copy it or reference the same diagram no need two write the same thing twice.

npetzall:
If you check your output folder diagrams and diagrams/orphans you will notice that table in orphans also exist in diagrams.

There are different diagrams in the table.html and orphans.html

RMsiemens:
@npetzall I just run it again and did only get one .png per table. Either one in diagrams/orphans or one in diagrams.  

image


I used following parameter:
```
    -t mysql -host localhost -port 3306 -db <db> -u root -p password -s <schema> -useSSL false -o 
    ./output -dp ./drivers/mysql-connector-java-6.0.6.jar
```

npetzall:
That's really strange, so if you got to tables/schema_version.html there is no diagram?

RMsiemens:
Yes, there is NO diagram.
On the tables with relationships there is a diagram in the "Relationships" section. But in the tables/schema_version.html the "Relationships" section is not there and no diagram is shown.

npetzall:
I wonder if we should change so that there always is a diagram for a table, and then in orphans page link to that tables 1degree diagram. Instead of having two different code paths to produce diagram. This would also avoid the corner case with implied creating a orphan diagram in the tables page.

RMsiemens:
Yes, this is maybe the better approach.  
PS: Do you have any Idee why I do not get the diagrams for the orphan Tables?

npetzall:
Regarding the PS why there isn't a diagram in the tables html page?

RMsiemens:
Yes exactly. You expected a .png file for each table in the "diagrams" folder. Any Idea what the problem could be that I do not have all tables?

npetzall:
https://github.com/schemaspy/schemaspy/blob/1ac8912d3559304e84f6ca7ae7d1a0e57346825e/src/main/java/org/schemaspy/view/HtmlTablePage.java#L184

I have one in the expecting due to implied relationship is seems to effect the maxChildren, maxParents.

If you have a truly orphan, no implied, it will not have a diagram. But I think that it should be changed so that there is always a visual representation of each table. 1st degree and no relationships should always have all details and then no writing of orphan diagram just find all orphans and link to the tables 1st degree diagram.

RMsiemens:
You are right. Removing the if statement here:
https://github.com/schemaspy/schemaspy/blob/1ac8912d3559304e84f6ca7ae7d1a0e57346825e/src/main/java/org/schemaspy/view/HtmlTablePage.java#L184

and here:
https://github.com/schemaspy/schemaspy/blob/1ac8912d3559304e84f6ca7ae7d1a0e57346825e/src/main/java/org/schemaspy/view/HtmlTablePage.java#L223

was the solution. (btw: DRY)

**Next Steps**
So how should we proceed?

npetzall:
We should se if we can force details when no children and no parent or have a look at how that is decided.
Then we should look at skipping the write orphans call, and just use the diagram from the table page. 

There might be some temporal coupling. Currently Orphans page is created before the individual table pages.

RMsiemens:
Actually removing just the two ifs mentioned above works quite well. The 'schema_version' diagram is now visible in the 'schema_version.html'

npetzall:
Should be small increments of changes.

npetzall:
Restarted there is an intermittent issue with mssql docker container.

RMsiemens:
Can this 'small increment of change' be merged to master or do I have to do any additional changes/work?

npetzall:
I'll merge once I get 30 seconds of time.

npetzall:
@RMsiemens Thank you for your contribution.

RMsiemens:
@npetzall Thank you. It was very pleasant working with you.

npetzall:
If you want you can continue to work with us to improve and modernize schemaspy. Moste code is between 11 and 4 years old.

We have ambition to create libraries and also create maven and gradle plugins.

PR #377, opened 2018-07-06
Fix sqlite.properties for xerial/sqlite-jdbc

0xBADDCAFE:
In currently, sqlite JDBC driver that actively maintained is [xerial/sqlite-jdbc](https://github.com/xerial/sqlite-jdbc) and it get information easily. In usage of [bitbucket page](https://bitbucket.org/xerial/sqlite-jdbc), specify driver `org.sqlite.JDBC` instead of `SQLite.JDBCDriver`.

The driver `SQLite.JDBCDriver` might be that [sqlitejdbc](https://github.com/crawshaw/sqlitejdbc) has. But it currently not maintained, can't refer [official page](http://www.zentus.com/sqlitejdbc), and difficulty to get information.

And the driver uses `connectionSpec` to sqlite database file path. Prefixed `/` causes missing the file to specify root of filesystem. So remove it.

### DIFF
--- a/src/main/resources/org/schemaspy/types/sqlite-xerial.properties
+++ b/src/main/resources/org/schemaspy/types/sqlite-xerial.properties
@@ -1,0 +1,15 @@
+#
+# see http://schemaspy.org/dbtypes.html
+# for configuration / customization details
+#
+
+description=SQLite
+
+connectionSpec=jdbc:sqlite:<db>
+db=path to database or :memory:
+
+driver=org.sqlite.JDBC
+
+# Sample path to the SQLite drivers.
+# Use -dp to override.
+driverPath=sqlite.jar


0xBADDCAFE:
References:
https://stackoverflow.com/questions/13869792/visualize-sqlite-database-schemaspy
https://stackoverflow.com/questions/8840990/schemaspy-on-mac-for-sqlite

npetzall:
Sorry for late reply, summer vacation and stuff.

Safest way would be to create a new instead of modifying an existing.

I just assume that this actually works or have worked in the past. Should have been verified when it first was added. But I'll admit this was ages ago and might only work with a specific driver version. 

So I'm just gonna verify that we aren't changing something that is actually working.

0xBADDCAFE:
OK, I'll revert to modify exist one and create new one like the name `sqlite-xerial.properties`. Or please speicfy the name if has more better.

npetzall:
Sounds great, I've only gotten around to Google the original, found it but xerial seems much better. I'll add a Changelog or What's new to the documentation.

I think `sqlite-xerial.properties` will be good.

When sorted they should endup next to each other.

0xBADDCAFE:
Fixed.

> I'll add a Changelog or What's new to the documentation.

Thanks, I'm just thinking that needs anywhere to documented about this too.

PR #381, opened 2018-07-12
Primarykeys contains null

npetzall:
This should replace #372 

Instead of having null pointer guards when used, null guard is placed when potentially null would be added to be able to extract as much information as possible in the case of null.


npetzall:
Leaving early tomorrow need to get some sleep. 95% done with test. Saw I've used reflection earlier will use the same approach.

rafalkasa:
@npetzall you should know that what are you doing in the project is complete awesome. I am fully in respect to you and your engagement. Thank you !

npetzall:
Tests added.

PR #390, opened 2018-07-30
Added Materialized views with definition.

npetzall:
* Added viewType
* Updated query to have union between pg_view and pg_matview
* Added test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,18 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.testcontainers</groupId>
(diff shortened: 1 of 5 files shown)


npetzall:
With the current changes there is no visible difference between a materialized view and a normal view.
Which might be OK.

I have seen in the code that when using JDBC we fetch the type but we don't pass it to the model (table/view). If we did this we could have the type presented in the XML/HTML report.

@rafalkasa Should I proceed to add TYPE to models or report or are we satisfied with the current solution.

npetzall:
We add them for now, if distinction is needed we might need to add it later.

Issue #431, opened 2018-09-28
Oracle: selectRowCountSql does not work


karlkauc:
I am working on a Oracle Database. 
But I get a lot of erros "ORA-00942: table or view does not exist". 

In the ora.properties file the selectRowCountSql is:
`selectRowCountSql=select table_rows row_count from information_schema.tables where table_name=:table `

Unfortunaly Oracle DB does not provide a Information Schema. 
See: [Wikipedia](https://en.wikipedia.org/wiki/Information_schema)

> RDBMSs which do not support information_schema include:
> 
> - Oracle Database
> - IBM DB2
> - SQLite
> - Sybase ASE

For Oracle this should work:
`selectRowCountSql=SELECT NUM_ROWS as row_count FROM ALL_TABLES WHERE TABLE_NAME = :table AND owner = :owner `

npetzall:
For now you can do a workaround by creating your own databaseType https://schemaspy.readthedocs.io/en/latest/started.html#create-your-own-db-type
So you can override the sql I'll have closer look when I get the time.

npetzall:
@karlkauc Just to check which version of SchemaSpy are you using?

We should fix this, but you shouldn't see any exception if your running 6.0.0 or 6.0.1-SNAPSHOT

karlkauc:
Thank you. 
I always run apps with -debug to see the log output. :smiley:

npetzall:
Always is a strong word. But then again you found something to improve.

npetzall:
Thanks for the PR.

PR #432, opened 2018-09-28
fixed selectRowCountSql for Oracle DB

karlkauc:
In the ora.properties file the selectRowCountSql is:
`selectRowCountSql=select table_rows row_count from information_schema.tables where table_name=:table`

Unfortunaly Oracle DB does not provide a Information Schema.
The ALL_TABLES metadata work fine for all Oracle DB.


npetzall:
Fixes #431

Django987:
Is that commit already included in the .jar-file, that can be downloaded?

npetzall:
@Django987 Yepp, there is a link to snapshot downloads in the readme. If you don't want to use a snapshot version. You can just copy the content of the file from github to file in your workdir named `mydbtype.properties` and use `-t mydbtype`.

Easiest is just to download latest/snapshot from the readmed/overview.

Issue #369, opened 2018-06-25
Procedures Documentation Oracle


JacBotin:
Does anyone have the property file code to generate the oracle procedure documentation (orathin) ?? (selectRoutinesSql and selectRoutineParametersSql for Oracle)??
Thanks.

npetzall:
Basics for docs regarding databaseType:
```
When metadata in JDBC isn’t cutting the mustard. You can replace it with a sql query. They are prepared and supports named parameters as long as they are available. Data is retrieved by column label. So additional columns are ok, but you might need to alias columns so that they are returned correctly to schemaspy.

:dbname
    DatabaseName -db
:schema
    Schema -s
:owner
    alias for :schema
:table
    table that the query relates to (think selectRowCountSql)
:view
    alias for :table
:catalog
    Catalog -cat
```

selectRoutinesSql
```
selectRoutinesSql=
    routine_name, routine_type, dtd_identifier, routine_body, routine_definition,sql_data_access, 
    security_type, is_deterministic, routine_comment
```
Above everything is string/varchar except is_deterministic which should be boolean.

selectRoutineParametersSql
```
selectRoutineParametersSql=
    specific_name, parameter_name, dtd_identifier, parameter_mode
```
Above everything should be string/varchar

And here is some Oracle documentation: https://docs.oracle.com/cd/B19306_01/server.102/b14237/statviews_2025.htm#i1585339
https://docs.oracle.com/cd/B19306_01/server.102/b14237/statviews_2063.htm#i1588578

There needs to be some aliasing and some joining to get it working. I'm not an oracle expert but I hope this gets someone started. If not I'll revisit this later.

JacBotin:
Hello!

It follows link of an excellent reference to generate the documentation of the oracle programs.
https://sourceforge.net/projects/ora-info-schema/?source=typ_redirect
;)

Sds

npetzall:
I'll have a look at it, as I understand it will create the information schema that exists for many others. With this information we should be able to create the required queries.

guicampos:
```Disclaimer: 
: You are responsible for your own environment. 
: Make sure to assure that your database user can really read DBA METADATA before executing this. 
: Always preffer to execute this in a non-production database to avoid IO contention or competition.
: This is working and in full use with Oracle Database 12c Standard Edition Release 12.1.0.2.0 - 64bit Production
```

I have made a simple configuration that gathers procedures, functions, triggers and packages information to the ROUTINES section of schemaspy.
[...]

Issue #329, opened 2018-05-17
MySQL 5.6: "Schema name can't be null" error


sampablokuper:
## Expected Behavior

```
$ ssh -f -o ExitOnForwardFailure=yes -L 10000:localhost:3306 \
        -i ~/.ssh/id_rsa_passwordless@192.168.122.68 sleep 10 ; \
    java -jar schemaspy-6.0.0-rc2.jar -t mysql \
        -db my_mysql_db -u my_mysql_user -p my_mysql_pwd \
        -host 127.0.0.01 -port 10000 -o output \
        -dp drivers/mysql-connector-java-8.0.11.jar
  ____       _                          ____
 / ___|  ___| |__   ___ _ __ ___   __ _/ ___| _ __  _   _
 \___ \ / __| '_ \ / _ \ '_ ` _ \ / _` \___ \| '_ \| | | |
  ___) | (__| | | |  __/ | | | | | (_| |___) | |_) | |_| |
 |____/ \___|_| |_|\___|_| |_| |_|\__,_|____/| .__/ \__, |
                                             |_|    |___/

                                              6.0.0-rc2

SchemaSpy generates an HTML representation of a database schema's relationships.

INFO  - Starting Main v6.0.0-rc2 on my_pc with PID 27186 (/home/sampablokuper/schemaspy/schemaspy-6.0.0-rc2.jar started by sampablokuper in /home/sampablokuper/schemaspy)
INFO  - No active profile set, falling back to default profiles: default
INFO  - Configuration file not found
INFO  - Starting schema analysis
INFO  - Using database properties:
INFO  -   [schemaspy-6.0.0-rc2.jar]/org/schemaspy/types/mysql.properties
Loading class `com.mysql.jdbc.Driver'. This is deprecated. The new driver class is `com.mysql.cj.jdbc.Driver'. The driver is automatically registered via the SPI and manual loading of the driver class is generally unnecessary.
INFO  - Connected to MySQL - 5.6.40
```
&hellip; and from there, SchemaSpy should successfully generate output for the `my_mysql_db` database.

## Current Behavior
```
$ ssh -f -o ExitOnForwardFailure=yes -L 10000:localhost:3306 \
        -i ~/.ssh/id_rsa_passwordless@192.168.122.68 sleep 10 ; \
    java -jar schemaspy-6.0.0-rc2.jar -t mysql \
        -db my_mysql_db -u my_mysql_user -p my_mysql_pwd \
[...]

Issue #300, opened 2018-04-10
[MySQL] -db argument should not be required when -all is present


hauntingEcho:
attempting to run against a MySQL database with the `-all` flag fails unless the `-db` flag is present.  As MySQL does not have a concept of multiple databases per server, the -db flag should not be required.

## Expected Behavior
* when `-all` and `-t mysql` are specified, the `-db` flag should not be required.
* if possible, the `-db` flag should not be required at all for `-t mysql` (just `-s` or `-all`)

## Current Behavior
`*** Required parameter 'db' (database name) was not specified.  It is required for this database type. ***`

## Steps to Reproduce (for bugs)
1. create local mysql database with user `testUser` identified by 'password'
2. pull the most recent schemaspy docker image
3. `docker run -v /Users/matt.beary/Desktop/example:/output schemaspy/schemaspy:snapshot -all -u testUser -p password -t mysql -host docker.for.mac.localhost -port 3306`

## Context
* Database type (e.g. MySQL): `mysql  Ver 14.14 Distrib 5.7.21, for osx10.13 (x86_64) using  EditLine wrapper`
* Database driver version (is you use a special one): from Docker image
* Your DDL (sample how to reproduce the issue): `create user testUser@localhost identified by 'password'`
* Full command including arguments how you run SchemaSpy: `docker run -v /Users/matt.beary/Desktop/example:/output schemaspy/schemaspy:snapshot -all -u testUser -p password -t mysql -host docker.for.mac.localhost -port 3306`

## Your Environment
* Version used: Docker :snapshot (March 26 seems to be the last update)
* Environment name and version (e.g. Chrome 39): Docker 18.03.0-ce-mac60 (23751)
* Operating System and version (e.g. Windows 10): OSX 10.13.4
* Java version: from Docker image

npetzall:
Well for the concept of multiple databases per server or instance. You have examples in the previous two issues where you use statement as create database on the same server.

Jdbc for mysql also has database as an argument for directly connecting to a database and not the dbms. 

But due to usage of -db spread all over the code. It's used in jdbc connection and in html report, probably xml as well. So it won't  be changed in near future, if possible.

hauntingEcho:
[...]

Issue #287, opened 2018-03-20
Failed to determine auto increment status


HezhiWang:
Hi, 

I encountered an issue when I run schemaspy to snowflake, here is log output:

SchemaSpy generates an HTML representation of a database schema's relationships.

INFO  - Starting Main on wanghezhis-MacBook-Pro.local with PID 37583 (/Users/wanghezhi/Desktop/Knotel/schemaspy/schemaspy-6.0.0-rc2.jar started by wanghezhi in /Users/wanghezhi/Desktop/Knotel/schemaspy)
INFO  - No active profile set, falling back to default profiles: default
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.springframework.cglib.core.ReflectUtils$1 (jar:file:/Users/wanghezhi/Desktop/Knotel/schemaspy/schemaspy-6.0.0-rc2.jar!/BOOT-INF/lib/spring-core-4.3.11.RELEASE.jar!/) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of org.springframework.cglib.core.ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
INFO  - Configuration file not found
INFO  - Starting schema analysis
INFO  - Using database properties:
INFO  -   /Users/wanghezhi/Desktop/Knotel/schemaspy/snowflake.properties
INFO  - Connected to Snowflake - 2.35.1 (driver change version: 3.5.3)
INFO  - Gathering schema details
Gathering schema details...WARN  - Failed to determine auto increment status: net.snowflake.client.jdbc.SnowflakeSQLException: Cannot perform SELECT. This session does not have a current database. Call 'USE DATABASE', or use a qualified name.
WARN  - SQL: select * from DBT_HWANG."ALL_DAYS" where 0 = 1
WARN  - Unable to extract the number of rows for table ALL_DAYS, using '-1'
WARN  - net.snowflake.client.jdbc.SnowflakeSQLException: Cannot perform SELECT. This session does not have a current database. Call 'USE DATABASE', or use a qualified name.
.WARN  - Failed to determine auto increment status: net.snowflake.client.jdbc.SnowflakeSQLException: Cannot perform SELECT. This session does not have a current database. Call 'USE DATABASE', or use a qualified name.
[...]

PR #434, opened 2018-10-10
if database name is not provided xml filename is null

karlkauc:
fallback to dbmsMeta.getProductName
fallback to hard coded "UNKNOWN"

### DIFF
--- a/src/main/resources/org/schemaspy/types/ora.properties
+++ b/src/main/resources/org/schemaspy/types/ora.properties
@@ -41,6 +41,12 @@ selectColumnCommentsSql=select table_name, column_name, comments from all_col_co
 #  thanks to Mikheil Kapanadze for the SQL
 selectRowCountSql=SELECT NUM_ROWS as row_count FROM ALL_TABLES WHERE TABLE_NAME = :table AND owner = :owner 
 
+# select any stored procedures and functions
+selectRoutinesSql=SELECT s.owner || '.' || s.name AS routine_name, s.TYPE AS routine_type, a.DATA_TYPE AS dtd_identifier, 'PL/SQL' AS routine_body, dbms_xmlgen.convert(xmlagg(xmlelement(e,s.text,'').extract('//text()') order by s.line).GetClobVal(), 1) AS routine_definition, null as is_deterministic, null AS sql_data_access, null AS security_type, null AS sql_mode, null AS routine_comment  FROM all_source s LEFT OUTER JOIN all_arguments a ON a.OWNER = s.owner AND a.OBJECT_NAME = s.NAME AND a.DATA_LEVEL = 0 AND a.argument_name IS null where s.owner = :schema GROUP BY s.owner, s.name, s.TYPE, a.data_type
+
+# select parameters for stored procedures and functions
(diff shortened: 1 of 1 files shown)


npetzall:
Sadly we currently have the same logic for name duplicated in HtmlMainPage or something similar with a link to the xml file. Same logic needs to be applied there until we have figured out a good Output Result object to contain the computed filename.

karlkauc:
I see.
What about extending the getName() in Database.java
```java
    public String getName() {
        if (Objects.nonNull(databaseName)) {
            return databaseName;
        }
        if (Objects.nonNull(dbmsMeta.getProductName())) {
            return dbmsMeta.getProductName();
        }
        return "UNKNOWN DATABASE NAME";
    }
```

npetzall:
How would you feel about introducing a OutputResult class with a list of Links that is just a simple object with href and name.

This would allow us to move some other logic to individual classes as well, the insert delete order for instance.

karlkauc:
Good idea. 

Database name could also be selected via the database metadata (enhancing the properties files). 

For Oracle this would be: 
```sql
select ora_database_name from dual;
```

For SQL92 compliant databases this would be: 
```sql
select SCHEMA_NAME from information_schema.schemata;
```

npetzall:
I'll write some more later. But if we think of something good we could do bigger changes for 7.0

Currently db (called databaseName) is used from the commandline for certain dbms when connecting and moste sql overrides currently assumes a selection of default database/schema from connectionstring.

So having a user supplied databaseName and a dbms supplied might be confusing.

npetzall:
So the problem is that database can be null, this is since `-db` is "required" but not enforced.
`-db` can be marked as required.

Looking over the database types we have. It's seems that they all have `<db>` in the connectionString.
From that point of view DB should be required. Since it's also required for xml and for html generation.

But removing `-db` has been mentioned #300 and skipping `-db` in connectionString causes #287 

But the usage could be removed. JDBC only gives catalog and schema. We could just remove the usage of Database.getName() and build the fileName with only getCatalog().getName() + "_" + getSchema.getName() + ".xml" (since Catalog is/should be the databaseName and schema i schema)

Then work could be continued on removing `-db` since I believe that SchemaSpy should have the behavior described in https://github.com/schemaspy/schemaspy/issues/329#issuecomment-391483518
when reading it remember catalog is database and schema is schema and we should ignore mysql until they have decided on how they wanna move forward in their jdbc-driver.

Instead of using user input in output we use dbms supplied data.

npetzall:
Last commit could you move that to a new PR and reference issue #369 in the commit?

npetzall:
Commit #301798 needs to be dropped from this PR.

And what do you say about changing file naming to [Catalog].[Schema].xml and also adding the OutputResults class?

karlkauc:
Hi,
yes [Catalog].[Schema].xml is a good idea.
Thanks

npetzall:
@karlkauc Will you do another PR or should I fixit after I've abstracted Graphviz?

karlkauc:
I will do another PR - but not today. 

I would like to add some features I still miss to use it in my office. 
E.g. triggers, table size, materialized views, user-defined data types, partitioned tables, ...

I hope I can do some PR in the next few weeks.

npetzall:
Could we create issue for these enhancements so that we can discuss implementations and also collaborate on solutions.

PR #435, opened 2018-10-15
added selectRoutinesSql and selectRoutineParametersSql for Oracle DB

karlkauc:
this should solve https://github.com/schemaspy/schemaspy/issues/369

As guicampos mentioned - Oracle sometimes is a little special.
Oracle can group functions and procedures in packages. So there are no
"package parameters".

I do not use dba_xxx views and DBMS_METADATA.GET_DDL function - because
you need special rights on the database to use it
(eg SELECT_CATALOG_ROLE).


npetzall:
@karlkauc Thank you, this is at least a first step. What I gathered from the discussion in #369 SchemaSpy model can't quite accommodate how Oracle has modeled Functions/Packages/Overloading.

Issue #328, opened 2018-05-17
Handle Awkward/Cryptic Table and Column Names - Feature Request


Gitarski:
I am requesting providing a way to handle odd/cryptic table and column names.  I'm sure other people know the following scenario.  You've taken over a project or given the task of making sense of a database.  The table and column names are all abbreviations and they are related to other tables and columns that have cryptic names as well.

For those looking to benefit from the SchemaSpy documentation, perhaps the diagrams it creates can be modified to use part of the comments to decorate the table and column names.  Especially useful in the relationship diagram.

I'm thinking it would that the first sentence of the comment or perhaps specifying a delimeter, such as "::", and anything before the delimiter will be added to the diagrams.  Having one or more flags to control the use of the tags in the diagrams would be useful.  For example, the table comment for table PPTID could be "PayPal Transaction ID::This table contains IDs for all PayPal transactions"". Column comments would be the same.

npetzall:
I know the feeling I've worked with arts which had lots of abbreviations.

I don't know yet to what extent tooltips can be used but would that be an option. With a large database adding more information in the diagram might just clutter it. While a tool tip could show information temporarily while hovering. It would also then be possible to just use the entire comment. 

Other options would be some kind of alternate name and have it processed using javascript/groovy snippet to give full control.

Gitarski:
The tooltips could be useful.  I think the main drawback of that is that if you're working on a SQL query and looking up the relationships it would be ideal to display the comment or tag all the time so that you don't have to keep moving the mouse and switching between windows.  Also, they wouldn't be useful if you wanted to print the diagram.

I was thinking of using the entire comment but I normally use more descriptive comments that tend to be lengthy and would clutter up the diagram as you said.

These features usually evolve as time progresses.  I tend to add a feature with simple functions in a specific configuration.  I then add or modify/improve them incrementally.

I was thinking an initial implementation would not have any configuration.  The first diagram could be the compact relationship diagram.  If the table or column comment has the tag delimiter of "::" then anything before the delimiter would be added to the diagram.  For tables, I would add it below the real table name.  For columns I'm not sure which is better, beside the name or below it.  I'm thinking beside the real column name so that tables with many columns stay more compact.

By using this tag in the comment, the documentation stays within the database.  If the user wants to use the tag, they add it. If they don't, it looks the same as before.  They can make the tag as long or short as they want.

npetzall:
If you could supply an example table, if that's doable/allowed from you database that would be great.
[...]

PR #453, opened 2018-11-05
Removed dependency to Mustache in DotNode

npetzall:
* Reduced complexity and duplication
* Later we might want to do this a lot cleaner

### DIFF
--- a/src/main/java/org/schemaspy/util/Html.java
+++ b/src/main/java/org/schemaspy/util/Html.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2018 Nils Petzaell
+ *
+ * This file is part of SchemaSpy.
+ *
+ * SchemaSpy is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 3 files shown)


rafalkasa:
Hi @npetzall for me the changes look good. 
Please consider using mustache to manage the DotNode templates in the same way as we have done with .html layout.

We will have xml files with DotNode layout and files will be generated by mustache engine. 
@npetzall what do you think about?

npetzall:
I've looked at a dot-lang library earlier.

The mustache thing would work. In reality it wouldn't be that horrible to have as a dependency for dot file generation. I'll think about it.

npetzall:
I'm gonna ponder this for a while, I think #328 is interesting.

I need to take a closer look on the width thingy and see what can be done. Probably need to create some dto to feed into the template. It's an interesting approach.

npetzall:
@rafalkasa you said i look ok. Can we add this and we can update change later.

rafalkasa:
I have added the new issue about the possibility of use Mustache for **dot** files. #458

PR #460, opened 2018-11-20
adding trigger feature to mysql

bhartwichae:


buildNumber.properties
```diff
@@ -1,3 +1,3 @@
 #maven.buildNumber.plugin properties file
```
npetzall: This file should be excluded from commits.

We want to remove it in the future from the repository.

npetzall:
I've skimmed the DatabaseMetaData and there is no way to get triggers via JDBC. It probably always have to be custom queries.

Now I only wonder if we should consider at least MSSQL, MYSQL, PGSQL and Oracle.
Sadly we have little knowledge of with which DBMS schemaspy is mostly used.

npetzall:
@bhartwichae Do you want me to take over?

krismorte:
I really like this project, I'm a database admninistrator I will dedicate somte time to study the project and continue this feature

PR #481, opened 2018-12-14
add new parameter relation depth degree 1 or 2 to simplify relation graphs

vdaburon:
I got some troubles with very big database (more than 3000 tables).
The output result is bigger > 4GB. The main cause is the 2sd degree relation graphs, some of graphs are > 11MB.
I add a new parameter relationdegreedepth,
You could choose the relation depth degree 1 or 2 (default value 2) to simplify relation graphs for database with lot of tables and relations.
The 1 value is best for big database with more than 1000 tables.

### DIFF
--- a/docs/source/new.rst
+++ b/docs/source/new.rst
@@ -28,6 +28,7 @@ What's new
 
 * Diagrams
     * Now has option to use embedded viz.js (no need for Graphviz) ``-vizjs``
+    * Choose the relation depth degree 1 or 2 to simplify relation graphs for database with lot of tables and relations
 
 * XML
     * Now includes routines
--- a/src/main/java/org/schemaspy/Config.java
+++ b/src/main/java/org/schemaspy/Config.java
@@ -72,6 +72,7 @@ public final class Config implements HtmlConfig, GraphvizConfig {
 
     private static final int DEFAULT_FONT_SIZE = 11;
     private static final int DEFAULT_TABLE_DETAILS_THRESHOLD = 300;
+    private static final int DEFAULT_RELATION_DEGREE_DEPTH = 2;
 
     private static final Pattern DBTYPE_PATTERN = Pattern.compile(".*org/schemaspy/types/(.*)\\.properties");
 
@@ -122,6 +123,7 @@ public final class Config implements HtmlConfig, GraphvizConfig {
     private String imageFormat;
     private String renderer;
     private Boolean paginationEnabled;
+    private Integer relationDegreeDepth;
     /**      
      * @deprecated replaced by -dp expanding folders
      */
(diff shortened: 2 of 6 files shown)


npetzall:
It might be a while before I have the time to look at this. It's a good feature.

docs/source/new.rst
```diff
@@ -28,6 +28,7 @@ What's new
 
 * Diagrams
     * Now has option to use embedded viz.js (no need for Graphviz) ``-vizjs``
+    * Choose the relation depth degree 1 or 2 to simplify relation graphs for database with lot of tables and relations
```
npetzall: Change to talk about degree of separation instead of depth

src/main/java/org/schemaspy/Config.java
```diff
@@ -626,6 +628,33 @@ public int getFontSize() {
 
         return fontSize;
     }
+    
+    /**
+     * @return
+     * @see #setRelationDegreeDepth(int)
+     */
+    public int getRelationDegreeDepth() {
```
npetzall: Would it be possible to move this to cli.CommandLineArguments

We actually want to remove the Config class.

It's not actually a big thing, since I personally think that we should switch from JCommander to Picocli since it's maintained. But that a whole other story.

src/main/java/org/schemaspy/output/html/mustache/diagrams/MustacheTableDiagramFactory.java
```diff
@@ -42,28 +42,38 @@
     private final MustacheDiagramFactory mustacheDiagramFactory;
     private final File outputDir;
     private final File tableDir;
+    private int relationDegreeDepth = 2;
```
npetzall: Final and not assigned

src/main/java/org/schemaspy/output/html/mustache/diagrams/MustacheTableDiagramFactory.java
```diff
@@ -42,28 +42,38 @@
...
+        
         File oneDegreeDotFile = new File(tableDir, table.getName() + ".1degree.dot");
-        File twoDegreesDotFile = new File(tableDir, table.getName() + ".2degrees.dot");
         File oneImpliedDotFile = new File(tableDir, table.getName() + ".implied1degrees.dot");
-        File twoImpliedDotFile = new File(tableDir, table.getName() + ".implied2degrees.dot");
+        
+        File twoDegreesDotFile = null;
+        File twoImpliedDotFile = null;
+        
+        if (relationDegreeDepth == 2) {
```
npetzall: Is this if block needed?

src/main/java/org/schemaspy/output/html/mustache/diagrams/MustacheTableDiagramFactory.java
```diff
@@ -77,38 +87,42 @@ public MustacheTableDiagramFactory(DotFormatter dotProducer, MustacheDiagramFact
...
-                }
-                if (sameWritten(oneImplied, twoImplied)) {
-                    Files.deleteIfExists(twoImpliedDotFile.toPath());
-                } else {
-                    MustacheTableDiagram twoImpliedDiagram = mustacheDiagramFactory.generateTableDiagram("Two implied", twoImpliedDotFile, table.getName() + ".implied2degrees");
-                    twoImpliedDiagram.setIsImplied(true);
-                    diagrams.add(twoImpliedDiagram);
-                }
-                return diagrams;
+            WriteStats oneImplied = new WriteStats(stats);
```
npetzall: This is not guarded
```
if (!impliedConstraints.isEmpty()) {
```
Probably what is causing the test failures.

src/main/java/org/schemaspy/output/html/mustache/diagrams/MustacheTableDiagramFactory.java
```diff
@@ -77,38 +87,42 @@ public MustacheTableDiagramFactory(DotFormatter dotProducer, MustacheDiagramFact
...
+            try (PrintWriter dotOut = Writers.newPrintWriter(oneImpliedDotFile)) {
+                dotProducer.writeAllRelationships(table, false, oneImplied, dotOut, outputDir);
             }
+            MustacheTableDiagram oneImpliedDiagram = mustacheDiagramFactory.generateTableDiagram("One implied", oneImpliedDotFile, table.getName() + ".implied1degrees");
+            oneImpliedDiagram.setIsImplied(true);
+            diagrams.add(oneImpliedDiagram);
+ 
+
+	        if (relationDegreeDepth == 2) {
```
npetzall: The ordering here will change the layout of the report.

1 degree:
1 real, 1 implied,
2 degrees:
1 real, 1 implied, 2 real, 2 implied

Can't say it's a huge issue, but it will change, grupping on degree instead of real and implied.

npetzall:
Superseeded by #485

vdaburon:
Thank to add this feature ... miss some documentation update to explain this new parameter

npetzall:
Have a look now, if you feel anything is missing, incorrect or confusing let me know.

I need to have a look at the web hooks tonight. Readthedocs hadn't been updated in 7 months. So added 6.1.0 and put it as default.

Issue #257, opened 2018-02-16
Tables  only defined in SchemaMeta produces nullpointer exception


npetzall:
## Expected Behaviour
Not a NullPointerException
Either ignoring the table or warning about it or just rendering it.

## Current Behaviour
NullPointer occurs since SchemaMeta Tables aren't registered with Markdown.java

## Possible Solution
Check first if table is added by SchemaMeta and ignore all processing of it. Diagrams and such.
Or register them with Markdown and have them rendered correctly.

## Steps to Reproduce (for bugs)
See issue #253

npetzall:
Adding possibility to use schemaMeta only as source for html generation is interesting.
An xslt to transform the XML output and you could create the xml on one system, transform the xml and generate html on another system.

PR #484, opened 2018-12-26
Db isolation

npetzall:
Starting isolation, only moving, creating interfaces will require much more work.

Due to the multischema and the `-dbhelp`


npetzall:
Move of markdown registration might solve #257 since tables added by schemameta will be registered. Haven't checked. Will check before merge aka tonight.

PR #491, opened 2019-01-03
Small fixes

vonloxley:
* Set default Sybase-driver to JConnect7
* Fix selectRowCountSql for Sybase
* Catch an unchecked SqlException


npetzall:
Hi, thanks for your contribution.

Would you mind sharing the exception that was thrown just curios. 

Also we have taken an approach to not break backwards compatability by adding new versions as new files and only patch existing versions of the database types. 

So could you just switch the database properties so that the newer has a version number and the old doesn't. Don't want unexpected changes.

PR #556, opened 2019-05-07
typo

karlkauc:
I think there is a typo - as there is no "-schemSpec" parameter.

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -64,7 +64,7 @@ Processing
 [-schemas listOfSchemas]
     List of schemas to analyze, separated by space or ``,`` or ``'`` or ``"``
 [-all]
-    Try to analyze all schemas in database, schemas can be excluded with ``-schemSpec`` which as defaults set by databaseType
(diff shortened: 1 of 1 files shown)


npetzall:
Might be a second issue, I'm unsure if the argument is case-sensitive, which means the [-schemaspec] is incorrect on the line below.

npetzall:
It's the definition of the option that is incorrect. So `-schemaSpec` is correct and [-schemaspec] is the typo.
Good catch, will fix.

npetzall:
It's the definition of the option that is incorrect. So `-schemaSpec` is correct and [-schemaspec] is the typo.
Good catch, will fix.

npetzall:
It's the definition of the option that is incorrect. So `-schemaSpec` is correct and [-schemaspec] is the typo.
Good catch, will fix.

npetzall:
It's the definition of the option that is incorrect. So `-schemaSpec` is correct and [-schemaspec] is the typo.
Good catch, will fix.

npetzall:
Replaced by #1093

PR #563, opened 2019-06-05
Initial support for DB2 for i (former AS/400)

francoiskha:
Hi,
I ran schemaspy-6.0.0 successfully on a DB2 for i V7R3m0 with the JDBC driver jtopen-9.8.
This is the configuration file.

PR #564, opened 2019-06-07
Adding routines support for DB2 for i (former AS/400)

francoiskha:
Hi,

I missed routines support in #563, so here it is.


npetzall:
Awesome, thanks

Issue #488, opened 2019-01-02
PostgreSQL v11 - column "Default" shows null even when column has no default defined


zam6ak:
On the table detail view, in the Columns grid, "Default" shows as "null" even if there is no default defined for the columnn.
Furthermore, even non nullable columns (e.g. "Nulls" shows no check mark) show "null" as "Default"

As an example, take a look at `public.film` table in sample DB (see details below)
Columns `title` and `language_id` are defined as `NOT NULL` and have no `DEFAULT` - but they show 'null' as default
Column `length` is NULLable and has no `DEFAULT` - also shows 'null' as default

If there is no Default defined for column, it should show up as empty/space

---
Tested using:

* DB:        PostgreSQL v11
* Driver:    [PostgreSQL JDBC 4.2 Driver, 42.2.5](https://jdbc.postgresql.org/download/postgresql-42.2.5.jar)
* Schemaspy: [schemaspy-6.1.0-SNAPSHOT](https://github.com/schemaspy/schemaspy/raw/gh-pages/schemaspy-6.1.0-SNAPSHOT.jar)
* Sample DB: [DVD Rental](http://www.postgresqltutorial.com/postgresql-sample-database/)
* Java ver:  1.8

npetzall:
I'll check the behavior of a cpl of dbms (mssql, mysql, oracle, pgsql, informix) and see how the jdbc drivers report default values.

npetzall:
Also check https://github.com/schemaspy/schemaspy/blob/a28c9fc932cc6f85c7780050a678b3a3d7f595e9/src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java#L81 since it probably needs handling when things are changed.

PR #574, opened 2019-07-27
include additional data in the schemaspy xml output

abdelgadir:
support for outputting a column's jdbcTypeCode and dbDefaultValue.
support for outputting sequences information - currently only for postgres.

note this additional information is NOT used when building the html output so the above info is NOT (yet) visible in the generated html gui.


npetzall:
Hi, thanks for contributing.

There are some minor things, I'll comment on them later.

I need to read up on Sequence in pgsql, it's pretty similar to autoincrement in some other databases and I wonder if there is a more generic model that can be used. I believe that we are missing autoincrement details for multiple databases. 

They might not be if a sequence can be shared by multiple tables. I think I've heard that it might be some kind of UDT.

abdelgadir:
Hi , sure , looking forward for your feedback.

In my opinion, auto increment is different and should be modelled separately from sequences. The current autoUpdatable attribute is all that is needed, I think, to describe auto increment columns (also known as identity in some databases). That is to say, as user I would only be interested in knowing that a column is auto incremented or not - how it is auto incremented will be a db implementation detail of no importance. 

The three databases I know of that support sequences: oracle, postgres and H2 , all share a very similar structure and can certainly be modelled uniformly. They all support a sequence name, start value, increment , min value and max value. I think also they all support caching properties that can also be modelled uniformly but I don't this is important to output.

Thanks

npetzall:
I looked it up and Sequence can be shared, however the model should also fit for autoincrement/identity columns, this can later be linked and looked up in the html report and/or add references in the xml. I'm a bit allergic to references, but from my point of view, sequences should probably go before tables.

So that you can read all sequences and when reading tables you can read them from cache instead of doing a second pass.

npetzall:
I see your point. I double checked the auto_increment and it doesn't need any additional information since it's always incremented by 1. (mysql)

From my point of view the interesting thing with Sequence is the increment since it can be something different to 1. This is the case for Identity in mssql, it has a seed and increment which fits nicely into a sequence which has a starting value and a specific increment.

buildNumber.properties
```diff
@@ -1,3 +1,3 @@
 #maven.buildNumber.plugin properties file
-#Tue Dec 12 15:37:44 CST 2017
-buildNumber=40
+#Wed May 01 22:38:17 BST 2019
+buildNumber=46
```
npetzall: This file should be removed from the pull request.
In the future we want to remove it.

src/main/java/org/schemaspy/input/dbms/service/SequenceService.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (C) 2019 Nils Petzaell
```
npetzall: Should be your name not mine =)

src/main/java/org/schemaspy/output/xml/dom/XmlProducerUsingDOM.java
```diff
@@ -96,6 +97,7 @@ public void generate(Database database, File outputDir) {
 
         xmlTableFormatter.appendTables(rootNode, tables);
         xmlRoutineFormatter.appendRoutines(rootNode, database.getRoutines());
+        xmlSequenceFormatter.appendSequences(rootNode, database.getSequences());
```
npetzall: I think i want this before tables or possibly before routines.
Since I believe that tables and/or routines might use Sequences.

src/main/java/org/schemaspy/output/xml/dom/XmlSequenceFormatter.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (C) 2018 Nils Petzaell
```
npetzall: It's your creation, so it's your copyright.

src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java
```diff
@@ -74,6 +76,9 @@ public Node appendColumn(Node tableNode, TableColumn column) {
         DOMUtil.appendAttribute(columnNode, "id", String.valueOf(column.getId()));
         DOMUtil.appendAttribute(columnNode, "name", column.getName());
         DOMUtil.appendAttribute(columnNode, "type", column.getTypeName());
+        DOMUtil.appendAttribute(columnNode, "typeCode", String.valueOf(column.getType() == null ?
+                DEFAULT_JDBC_TYPE_CODE : column.getType()));
+        DOMUtil.appendAttribute(columnNode, "defaultValue", String.valueOf(column.getDefaultValue()));
```
npetzall: Just for your information.
#488 
We have an issue that, we might report string "null" when it's an actuall null.

buildNumber.properties
abdelgadir: ok will revert it.

src/main/java/org/schemaspy/input/dbms/service/SequenceService.java
abdelgadir: sure :)

src/main/java/org/schemaspy/output/xml/dom/XmlProducerUsingDOM.java
abdelgadir: ok

src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java
abdelgadir: i see yes string.valueof does that

npetzall:
@abdelgadir If I could trouble you for a squash that would be neat. Then I'll merge.

abdelgadir:
hi @npetzall 
i am not sure how to do a squash. i thought you choose this option during the merge. see screenshot below 

image

npetzall:
I'll do it from the commandlin. I researched the github function and it has some issues.

Kid woke up so I'll do it in an hr or so.

npetzall:
I'll create a new PR with your changes and a few commits of my own to fix some tests that are failing.
It seem like travis-ci checks aren't working for us at the moment.

src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java
npetzall: defaultValue is now added twice.

https://github.com/schemaspy/schemaspy/blob/7595bc318bcb280629432052c21136e206e44051/src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java#L81

https://github.com/schemaspy/schemaspy/blob/7595bc318bcb280629432052c21136e206e44051/src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java#L86-L95

abdelgadir:
fixed the issue of duplicate defaultValue attribute output. The problem i had was that the defaultValue is only output if its value is not null (and i needed to be output always even if null). I have removed the line i added that outputs it. I guess for my requirement, i can go on the assumption that if the attribute itself is missing then its value is assumed to be null.

abdelgadir:
yeah i noticed that test maven goal works, but verify fails.

npetzall:
#582 will replace this pull request.
As I mentioned #488 there is some issues with the default value.

If "null" is the default or if null actually means that it's undefined.
When fixing the tests I got rows that looked similar to:
```
<column autoUpdated="false" defaultValue="null" digits="0" id="2" name="name" nullable="false" remarks="UserName" size="16" type="VARCHAR" typeCode="12"/>
```
The row is not nullable but the defaultValue is null.

It's too late at night right now, I'm about to fall a sleep.
But I think that doing testing with mysql, mssql, pgsql, oracle I could probably figure out how to handle the defaultValue, so that we can have "null" for actual null and maybe "undefined" if not set at all. Since it will fallback to dbms specific behavior.

abdelgadir:
hi @npetzall what is the state of this pull request please - is it going to be accepted or rejected? is there anything else i need to correct ?

npetzall:
Changes merged with #582, #488 is still an issue.

npetzall:
@abdelgadir I haven't had the time to check on #488 to get a handle on the defaultValue="null" but merged with the defaultValue="null" and handle the issue with #448.

abdelgadir:
many thanks @npetzall

PR #577, opened 2019-08-02
Sort output

rout39574:
Heya.  Thanks for the awesome tool!

I wanted to check your generated site into a repository.   Having the generated pages be stable across builds helps keep chaff down in the changes, and makes actual drift much easier to identify.

I'm no java coder, but I think I made some reasonably idiomatic suggestions here, and it Works For Me so far.

Would you consider this or similar changes?

This is my first stab at a PR; if I've blown the etiquette somehow, please let me know and I'll try to go back and do it right. :)

### DIFF
--- a/src/main/java/org/schemaspy/view/HtmlColumnsPage.java
+++ b/src/main/java/org/schemaspy/view/HtmlColumnsPage.java
@@ -37,6 +37,8 @@ import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TreeSet;
+import java.util.Comparator;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


npetzall:
Only thing I directly can think of is that we might need to add sorting in more places.

This also seems to be the most natural sorting. So I think we should implement Comparable on MustacheTableColumn and simply add sorted() to the stream and finally collected it as a LinkedHashSet.

But now you've made me go down the rabbit hole. MustachTableColumn is just a decorator class for TableColumn. Which has a private innerclass named ColumnComparator.
ColumnComparator is only used once, and as stated before the most natural ordering of TableColumns is by TableName and ColumnName.

So really TableColumn should implement Comparable thus removing the ColumnComparator and it's usage.
MustachTableColumn should implement Comparable and delegate to TableColumn.

Give it a shot, and if you feel it's to much let me know.
In short TableColum should implement Comparable, compareTo should have same logic as it's inner class ColumComparator. ColumComparator should be removed.
MustacheTableColumn should implement Comparable and compareTo should delegate to TableColumn.
Stream in HtmlColumnPage should include .sorted() and finally collected as a LinkedHashSet.

rout39574:
Aight, I'll give it a shot.

- Allen S. Rout

PR #581, opened 2019-08-05
Updated overview.rst because of a falls spelling

Kamyabnz:
I am working with SchemaSpy for a while now and went through the documentation a lot and I found a misspelling on the first page.

It should have been 'vizjs' not 'visjz'.


npetzall:
Great catch thanks.

PR #582, opened 2019-08-06
Pr547 additional data xml

npetzall:


npetzall:
As I mentioned #488 there is some issues with the default value.

If "null" is the default or if null actually means that it's undefined.
When fixing the tests I got rows that looked similar to:
```
<column autoUpdated="false" defaultValue="null" digits="0" id="2" name="name" nullable="false" remarks="UserName" size="16" type="VARCHAR" typeCode="12"/>
```
The row is not nullable but the defaultValue is null.

It's too late at night right now, I'm about to fall a sleep.
But I think that doing testing with mysql, mssql, pgsql, oracle I could probably figure out how to handle the defaultValue, so that we can have "null" for actual null and maybe "undefined" if not set at all. Since it will fallback to dbms specific behavior.

npetzall:
Left line: https://github.com/schemaspy/schemaspy/blob/a28c9fc932cc6f85c7780050a678b3a3d7f595e9/src/main/java/org/schemaspy/output/xml/dom/XmlColumnFormatter.java#L81 which causes the defaultValue="null" when the value is null. I'll add a comment about this in #488 just so that we know that it should be handled with the issue.

PR #585, opened 2019-08-15
commandline.rst updated - argument doesnt exist

Kamyabnz:
I am working with informix servers and tried to implement the '-noschema' argument but the Tag is not even used in Code so it always gives the same Error. In the Documentation I wanted to remove the sentence referring to '-noschema' argument. The same in 6.0.0 Documentation.

I hope we can improve the documentation alot more, because it would make my life much easier :)


npetzall:
It was removed in 872cf4c202be9fe930c477672fbfd14025d3bb6b

PR #590, opened 2019-09-12
Dockerfile: Update JRE and remove `java` user

abitrolly:
`java` user creates problems when mounting volumes.

See https://github.com/containers/libpod/issues/3990


abitrolly:
I checked several popular projects on DockerHub - all of them run as daemons with communication over network ports, and don't need volume mapping. All these projects use custom USER directive. I don't think that this PR solves the problem with filesystem permissions anymore. It should be addressed on `podman`/`docker` side. The relevant `docker` issue is https://github.com/moby/moby/issues/2259

abitrolly:
I am closing this issue to avoid sacrificing security for `docker` users. I used `podman cp <containerid>:/output/_data ./html`  to get report from the `schemaspy` run.

abitrolly:
For tested Java version - I didn't expect that minor Java updates can break applications, and I decided that build is enough. Or course, if such problem with Java is likely, and Java in the Docker image is not the same, then the image should be tested the same way as non-packed binary.

The version of Java for non-docker build is 1.8.0_151 https://travis-ci.org/schemaspy/schemaspy/jobs/584390999#L372. Looks old.

npetzall:
I can't swear it was only the minor Java update, since it's a packaged image alpine+Java the problem might have been with alpine. Yepp, all the Java stuff is getting old as in unsupported. So both compile and test should be bumped to Java 12. But there might be dragons.

abitrolly:
Yes, build on Java 12 fails - https://travis-ci.org/schemaspy/schemaspy/builds/584550410

I am not sure what exactly is missing. Looks like a problem with build system.

npetzall:
Seems to be an issue with jaxb and failsafe plugin in maven. So the plugin might need to be bumped or addition dependencies added for the plugin.

abitrolly:
Updating the plugin to the latest version magically fixed the build - https://github.com/schemaspy/schemaspy/pull/591

abitrolly:
I need to get back to it. Two reasons:

1. PostreSQL container doesn't use USER https://github.com/docker-library/postgres/blob/master/11/Dockerfile and it is far more popular (not far more secure)

2. People who care about security should run containers unpriviliged - use `podman` or `--user` flag (if that helps)

The primary use case for running SchemaSpy in a container is a distribution method, not security. Running `SchemaSpy` even under non-root under the same user already a greater security risk than broken root container.

npetzall:
The only defense I actually have is that it's in the docker best practices to run as an unprivileged user if root isn't required to run the application/service. Which is not needed.

I had a closer look at the Postgres you linked, it doesn't have the user directive, but Postgres doesn't run as root, it runs as postgres user if nothing else is specified. It has the same issues schemaspy has. Just read the documentation for it on docker-hub.

npetzall:
So the real issue might be that we don't fully support the --user argument. Which should possibly maybe solve the podman issue, since you could ask it to run with your user, or same uid. Which would grant it access the same as the user, which should be belove root, unless you run your system as root which would be crazy.

abitrolly:
This problem even got an article from RedHat - https://www.redhat.com/sysadmin/rootless-podman-makes-sense.

Running with my user bears that same security risks than running with root or without container - if the process escapes container, it can steal my keys.

Now the cost of the protection. The primary function for SchemaSpy is to produce the `/output` folder - without it it useless. The cost of getting the output is too high compared to the risk of actually shipping the SchemaSpy with malicious code that is also capable of breaking the container. That's why I would still remove USER.

Now I need to see what is this `podman unshare` - the recommended way to deal with USER according to the RedHat article.

npetzall:
https://github.com/containers/libpod/issues/4016#issuecomment-564850604
The example of postgres in the comment, I'll guess you would see sort of the same for schemaspy. (I'm not a podman user, so I'm not able to check)
Instead of using the USER directive they use gosu to drop privileges. So that the process runs with a different UID. We could do that, but it would mean we need to add gosu to the image to get the same result as USER directive I think.

What we could do which would make life easier is to specify and document the UID used inside of the container. Which would make the `podman unshare` command much simpler.

abitrolly:
Tthis `podman unshare` - is this an equivalent to `su <user>`?

abitrolly:
I don't have much time for this. unfortunately. I will use the old hack to extract the data from exited container.
```
podman run --name <containerid> ...
podman cp <containerid>:/output/_data ./html
podman rm <containerid>
```

npetzall:
Did you read the article you posted?
https://www.redhat.com/sysadmin/rootless-podman-makes-sense

If your hung up about postegres instead of schemaspy. Just replace them, the article is about host-mounting to a non-root UID running in a container. The solution applies to schemaspy as much as it does to postgres when you want the data to be persisted outside of docker.

abitrolly:
I read the article, but I miss some background on how kernel namespaces work, and without this the article is useless for me.

PR #594, opened 2019-09-16
Query for Routines/Procedures for an Informix DB and Solution

Kamyabnz:
I have been working on getting Routines out of an Informix DB, I do have a solution to get everything necessary but it needs changes inside the RoutineService.java but this solution just works with Informix DB and so i just given the SQL query that should work if the individual changes the getString(""); with Query Column names. 

- data Column gives just a part of the Procedure source code back so that is why i have the solution which we could talk about.

### DIFF
--- a/src/main/java/org/schemaspy/input/dbms/service/RoutineService.java
+++ b/src/main/java/org/schemaspy/input/dbms/service/RoutineService.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2019 Nils Petzaell
+ * Copyright (C) 2019 Kamyab Nazari
  *
  * This file is part of SchemaSpy.
  *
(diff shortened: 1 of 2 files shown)


npetzall:
Create a PR with all required changes. So that I can get a complete understanding of the problem.

Kamyabnz:
I have added the next Commit with the changes to the RoutineService,java, the code is not the most polished but it does the job for now. You could tell me what we can do with it, because at the en

Kamyabnz:
I think i have to commit it at ones so it works right ?, because it does the test on one commit and that is why it does not have the Query commit for the Informix.properties ? @npetzall

npetzall:
It will fail due to the changes in RoutineService.java.

There are a lot of changes that aren't needed. As examples. 

Procname should be aliased with as to routine_name.
isproc should be a switch or case and aliased to routine_type.

So with aliases and case we can probably fix most of them. I haven't got the time to look them all over at the moment so currently I'm not sure if any changes are needed in RoutineService at all.

Kamyabnz:
alright, we could do that. Most of the functions are to get the right and full Definition from the Query because, in case of Informix we have a situation where Query gives many rows of results about one Routine_Name and so while running through ResultSet it will override the Routine Details of that specific Routine_Name. So we have the details in the last row of the ResultSet with that Routine_Name. Informix gives for the Data/Definition just part of the Source Code because it can only carry 256 Characters, so I am just putting them Lists and while we start going through the ResultSet it stitches them toghether and puts them into the RoutineMap.

Thank you for the response, lets see what we can do with it.

npetzall:
If the only really needed change is to stitch the body. That could probably be implemented without breaking the other dbms.

Kamyabnz:
Well yes that is the actually the only thing but the ouputs are also not really understanable, because it give out for 'Type' = t out and you dont know what that is if you do not have the IBM documentation to tell you "t" means "PROCEDURE".

npetzall:
That's where you use the CASE statement so something like

```
SELECT
   CASE
      WHEN type = "t" THEN "Procedure"
      When type = "f" THEN "Function"
   END
   FROM routines; 
```

Just an example, but I might have a few hrs to look at it tonight.

Kamyabnz:
Alright i understand what you mean. Yes that would be nice

npetzall:
Won't contain the everything in one single page but should be good implementation of routines in ##653

Kamyabnz:
Then I could close this, should we still maybe take the stitching part of the code and add it into an issue where we take the idea and maybe later we implement also Stichling if it is an informix dB ?

npetzall:
It's currently implemented by appending data when lines have the same identifier. I hade to make the routine name include the parameter types so that they became unique.

Kamyabnz:
Ah i see what you did there, alright then we don’t need this solution anymore, I will close it.

Issue #586, opened 2019-08-15
-vizjs produces squished diagram titles


Kamyabnz:
Trying to use -vizjs for creation of diagrams, only relationships Diagrams look squished. By using the Master and the prebuild .jar

## Expected Behavior
The relationship Diagrams should look decent, the title should not be interfering with [Type]

## Current Behavior
The End of Diagram Titles are interfiering with table [Type] and you cannot read the last part of the title. With all of the Diagrams this is being seen and with longer titles it looks much worse. 

![Issue-Vizjs-piuc01](https://user-images.githubusercontent.com/32940190/63087682-0fd02200-bf53-11e9-81eb-73b923dfef94.PNG)

## Possible Solution
Maybe the VizJSDot.java should get some kind of title size reduction? 
To be honest, I don't know what you can do to make the diagram bigger and let the font stay the same size. I tried to make the title font smaller, but there is only one main font size option that sets of both texts the font size.

## Context
* Database type (e.g. MySQL): Informix
* Database driver version (is you use a special one): JDBC.4.10.JC4DE
* Your DDL (sample how to reproduce the issue):
* Full command including arguments how you run SchemaSpy:

x <--- Because because of privacy reasons

1. -jar schemaspy-6.1.0-SNAPSHOT.jar
2. -debug
3. -t informix 
4. -dp  D:\informix-drivers\JDBC.4.10.JC4DE\lib\ifxjdbc.jar;D:\informix-drivers\JDBC.4.10.JC4DE\lib\ifxjdbcx.jar
5. -host xxxx.xxxx.com
6. -port xxxx
7. -db axxxx
8. -server db
9. -connprops IFX_TRIMTRAILINGSPACES\\=1
10. -u username
11. -p password
12. -all
13. -vizjs
14. -o D:\OutputSchemaSpy6-1-Pre-made

## Your Environment
* Version used: Prebuild 6.1.0-SNAPSHOT and also latest Master maven build
* Environment name and version (e.g. Chrome 39): Chrome 76
* Operating System and version (e.g. Windows 10): Windows 10
* Java version: java version "1.8.0_212"

npetzall:
I'm gonna see if I can recreate this. I had a quick glance at code that produces the dot-file/content.
It' seems that the "table/box" width is only calculated for columns.
So if this is correct, having a table with long name and all columns short should produce the issue.

In either case, what would you prefer larger "table/boxes" or smaller text?

Kamyabnz:
[...]

PR #598, opened 2019-09-18
Added the dynamic MaxTitleWidth into DotNode.java for Title

Kamyabnz:
Because of small columns and big titles, the titles would have gotten squished, i have added the same procedure to the Titles too. The issue was: #586

### DIFF
--- a/src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java
+++ b/src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java
@@ -102,6 +102,7 @@ public class DotNode {
         String tableName = table.getName();
         // fully qualified table name (optionally prefixed with schema)
         String fqTableName = (table.isRemote() ? table.getContainer() + "." : "") + tableName;
+        int maxTitleWidth = getTitleMaxWidth(fqTableName);
         String colspanHeader = config.showColumnDetails ? "COLSPAN=\"4\" " : "COLSPAN=\"3\" ";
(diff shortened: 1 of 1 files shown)


Kamyabnz:
I found the problem, why we are getting this errors. The problem is with graphviz, that gives out the ERROR of the width being too small and so it tries to change it. But it still gives out errors so the build fails, I think we just have to check before that if graphviz is running or -vizjs has been choosen.

Kamyabnz:
Yes it should have and also if you try the same code with graphviz it gives out for example this error:
```
ERROR - D:\graphviz-2.38-Mid\release\bin\dot -Tpng produkt.implied1degrees.dot -oprodukt.implied1degrees.png -Tcmapx: Warning: cell size too small for content
ERROR - D:\graphviz-2.38-Mid\release\bin\dot -Tpng produkt.implied1degrees.dot -oprodukt.implied1degrees.png -Tcmapx: in label of node produkt
```

But the result does not have any problems. With both vizjs and graphviz in my case atleast.

npetzall:
Yeah should probably be a warning but we just read stderror so that's why it's classified as an error. I've seen the output before nothing new.

Kamyabnz:
Yes it should actually be a warning because it also says i the message but well as you did say it is classified as an error.

npetzall:
Did you recreate your fork or did I do something really horrible, like delete it  (which I shouldn't be able to). The branch for this PR is gone.

npetzall:
Closed with #615

PR #601, opened 2019-09-18
Clean up docker

### LINKED ISSUE #583: Jar inside of docker image
I'm currently working on an automation inside of our CI system and I would like to use schemaspy to generate documentation of the database.
I have done some modifications in the image, as I jenkins doesn't like when the image has an entrypoint.
Now I'll have to run an the image by doing the `java -jar schemaspy...jar`, however I see that the jar name has the version in its name.
It would help a lot if the jar inside of the image didn't contain version strings so that one can always refer to the jar with same name regardless of the tag of the image.
Of course I can do that by myself, but would be nice if this change was included in the official docker image. If I have time to do it, I'll send a pull request to change.

npetzall:
* Possible to override /output with ENV: OUTPUT
* Possible to override /drivers with ENV: DRIVERS
* Removed volumes

Fixes #583

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,4 @@
 *
 !docker/open-sans.tar.gz
-!docker/entrypoint.sh
+!docker/schemaspy.sh
 !target/schema*.jar
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
       - graphviz
 script:
   - ./travis.sh
-  - ./docker.sh build
+  - ./docker.sh build snapshot
 cache:
   directories:
     - '$HOME/.m2/repository'
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,20 +30,19 @@ RUN adduser java -h / -D && \
     curl -JLO http://search.maven.org/remotecontent?filepath=org/mariadb/jdbc/mariadb-java-client/$MARIADB_VERSION/mariadb-java-client-$MARIADB_VERSION.jar && \
     curl -JLO http://search.maven.org/remotecontent?filepath=org/postgresql/postgresql/$POSTGRESQL_VERSION.jre7/postgresql-$POSTGRESQL_VERSION.jre7.jar && \
     curl -JLO http://search.maven.org/remotecontent?filepath=net/sourceforge/jtds/jtds/$JTDS_VERSION/jtds-$JTDS_VERSION.jar && \
-    mkdir /drivers && \
     mkdir /output && \
     chown -R java /drivers_inc && \
-    chown -R java /drivers && \
     chown -R java /output && \
     apk del curl
 
 
-ADD target/schema*.jar /
-ADD docker/entrypoint.sh /
+ADD target/schema*.jar /usr/local/lib/schemaspy/
+ADD docker/schemaspy.sh /usr/local/bin/schemaspy
 
 USER java
 WORKDIR /
 
-VOLUME /drivers /output
+ENV DRIVERS=/drivers
+ENV OUTPUT=/output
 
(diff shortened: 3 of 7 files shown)


renannprado:
@npetzall I remember I made a comment here about variable names of `OUTPUT` and `DRIVERS` which IMO should be prefixed with `SCHEMASPY_` same as it's done for other containers such as postgresql due to variable clash issues in some environments. If you are willing to take this change, I create a pull request for it.

npetzall:
Sorry to bring this up as a discussion, but isn't the point with docker to avoid the clash since the container has its own envs, it's isolated and from way back when I learnt docker basics there was this golden rule single process in the container so there would be nothing to clash with. I'm not saying no, but I'm not saying yes either at the moment.

renannprado:
One process per container no questions asked, we're on the same page there. And yes this will most (if not always) give you the isolation you're talking about.

The thing is that sometimes you are running in environment (like jenkins) where all environment variables will be passed to the container while trying to run it and therefore it could be that an env var created by the pipeline is used by the container while it shouldn't ('cause they use these common/non-prefixed variable names). I'm not sure if there are more reasons why it's common to prefix environment variable names in docker but I'll mention a few examples:

https://github.com/docker-library/postgres/blob/master/9.4/docker-entrypoint.sh#L183
https://github.com/docker-library/php/blob/master/7.3/buster/apache/Dockerfile#L52
https://github.com/docker-library/mongo/blob/master/4.2/docker-entrypoint.sh#L193
https://github.com/docker-library/mysql/blob/master/8.0/docker-entrypoint.sh#L173
https://github.com/neo4j/docker-neo4j-publish/blob/master/3.5.9/community/docker-entrypoint.sh#L200

I could keep going. 

So as you can see many example (even more inside of the official images that dockerhub maintains) the variables are prefixed. Again, probably with reasoning beyond the use case of jenkins, but as you can see it's a common practice.

If you're still not convinced, I rest my case and will maintain the changes by myself somehow.

Thanks for your time.

npetzall:
I'll change this after kids has gone to sleep. It's not so much about convincing me about the change I wanted more of a reasoning behind prefixing in docker.

Lots of people do prefixing very often when applications are used in a shared environment. So it's nothing strange as such. 

I'll have to check the jenkins thing it sounds like a security vulnerability or at least something you need to think long and hard about. If it has env leakage. I mean one container could read them all and then post them somewhere. 

But normal usecase won't change with prefixing, and I guess few people use it. So we'll sneak in a breaking change.

renannprado:
> I'll have to check the jenkins thing it sounds like a security vulnerability or at least something you need to think long and hard about.

This is kinda a loophole, if someone is running code (or container) they don't trust, they should have an outgoing proxy through which they can whitelist only remotes that they trust. For example, even if I pass only database credentials to a container (and I do this for Flyway, Schemaspy and cloudsql proxy), how can anybody guarantee that these applications are not shipping my credentials anywhere else?

> So we'll sneak in a breaking change.

Is it really a breaking change? These variables were not present before, were they? 'cause schemaspy's `6.1.0` docker image (latest stable) doesn't yet support the environment variable.

npetzall:
Your build environment should never directly connect to the internet, you should have binary repo for any needed tools or dependencies and if need be use a proxy with a whitelist like you say.

We could assume that if you don't specify that the envs should be set for these containers that they will also be set for that container that runs html2pdf and you should be aware that it will see those envs, not obvious, but with that logic. But it totally destroys the isolation.

Nice catch, it's an unrelease minor, so we just correct the minor.

npetzall:
I had forgotten to update docker-hub so you might be the only who knows about the feature.

PR #603, opened 2019-09-19
Always produce table diagram

### LINKED ISSUE #376: Always show a diagram/visual representation of table
## Expected Behavior

All table pages has a diagram/visual representation of the database.

## Current Behavior

Orphan tables doesn't have it, except if they have implied which is a bit awkward.

## Possible Solution

Remove the condition that eliminates orphan tables.
In the long run we could remove the orphans diagram and just link/reuse the table diagrams.

npetzall:
fixes #376


npetzall:
Will also create diagrams for views. If database contains many view this might have a performance impact, but I sort of like the diagram. So unless it becomes an issue they will be created.

PR #629, opened 2019-12-03
modify selectPrimaryKeysSql

mapausch:
Modify selectPrimaryKeysSql to return a correct result when >1 indexes exist for a table.
Affects schemaspy#628 and schemaspy#546


npetzall:
https://travis-ci.org/schemaspy/schemaspy/builds/620330018 is OK.

PR #630, opened 2019-12-12
Allow to run other commands in in container

abitrolly:
If the first argument passed to entrypoint doesn't start with dash, it is the command that needs to be executed.

Taken from the best practices on ENTRYPOINT
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint

### DIFF
--- a/docker/schemaspy.sh
+++ b/docker/schemaspy.sh
@@ -1,15 +1,18 @@
 #!/bin/sh
 
-function run_schemaspy() {
+run_schemaspy() {
     [ -d $SCHEMASPY_DRIVERS ] && export DRIVER_PATH=$SCHEMASPY_DRIVERS || export DRIVER_PATH=/drivers_inc/
     echo -n "Using drivers:"
     ls -Ax $DRIVER_PATH | sed -e 's/  */, /g'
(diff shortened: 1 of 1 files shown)


npetzall:
Can you explain the need to run other commands in the container, so that I may better understand your needs. 

You can always make use of `- -entrypoint`

abitrolly:
Troubleshoot problems with executing SchemaSpy. Getting files out, inspecting the contents etc. For example making this PR work. I used to how PostgreSQL like containers allow to do this.

abitrolly:
I just found out that I was using a very old version of snapshot container with `/entrypoint.sh` and all this time I was trying to understand why the entrypoint was wrong.
```
podman inspect docker.io/schemaspy/schemaspy:snapshot
```

abitrolly:
I tested that the script works.

npetzall:
If it's used for troubleshooting `podman run --entrypoint=sh` should be used instead of building in support for arbitrary commands in the container.

abitrolly:
Then why `postgres` doesn't follow this advice?

npetzall:
What advice? The one that I gave specifically for SchemaSpy?

I'm not a maintainer/developer of postgres. But I'm gonna guess and extrapolate a lot from the documentation.

Postgres container contains both the dbms and the commandline client psql. Which means that it has more than 1 application that is targeted as such they must support both.

So you're more or less comparing two different scenarios. We don't need that support, they need it.

PR #679, opened 2020-05-13
link added for Rails-based naming convention

aisbergde:
additional one "/" is escaped because this prevent website simplifier to read the whole site

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -20,7 +20,7 @@ General
 [-configFile filePath]
     Path to configFile to be used, default is to look for schemaspy.properties
 [-o outputDirectory]
-    Directory to write the generated HTML/graphs to
+    Directory to write the generated HTML\/graphs to
 
 Database related
 ----------------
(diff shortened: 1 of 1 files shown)


docs/source/configuration/commandline.rst
```diff
@@ -94,7 +94,7 @@ Html report related
 [-nopages]
     Just list data as one long list instead of pages.
 [-rails]
-    Use Rails-based naming convention to find relationships between logical foreign keys and primary keys.
+    Use `Rails-based naming convention <https://gist.github.com/iangreenleaf/b206d09c587e8fc6399e#relations-in-models>`_ to find relationships between logical foreign keys and primary keys.
```
npetzall, line 97: I wonder if we should find a better example than a gist.

aisbergde: You send me the link, I only added to the documentation. Maybe the current implementation should be checked. If the only implemented thing is that a relation is build between
AnyTable.Profiles_ID is linked to Profiles.ID
Then it would be better to write this example

PR #689, opened 2020-05-15
-meta fileOrPath description extended

aisbergde:
I tried to use the option and I think the description could be a little bit extended

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -81,11 +81,9 @@ Processing
 
 Additional data
 ~~~~~~~~~~~~~~~
-[-meta fileOrPath]
-    Single schema analysis file path to one (1) SchemaMeta-xml-file;
-    when running ``-all`` or ``-schemas`` it defines the path to directory containing one ore more SchemaMeta-xml-files with pattern (DatabaseName|Schema).meta.xml
-    
-    For example, if you want to add metadata for 2 different schemas you need 2 files in this path: schema1.meta.xml, schema2.meta.xml
+[-meta pathToFolder]
+    Augment the database by using :ref:`schemameta`. Specifying path to a folder containing files named [schema].meta.xml, if schema is not used it will look for [databaseName].meta.xml |br|
+    One file per schema.
 
 Html report related
 -------------------
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -82,8 +82,10 @@ Processing
 Additional data
 ~~~~~~~~~~~~~~~
 [-meta fileOrPath]
-    Single schema analysis file path to SchemaMeta-xml, when running ``-all`` or ``-schemas`` path to directory containing
-    SchemaMeta-xmls with pattern (DatabaseName|Schema).meta.xml
+    Single schema analysis file path to one (1) SchemaMeta-xml-file;
+    when running ``-all`` or ``-schemas`` it defines the path to directory containing one ore more SchemaMeta-xml-files with pattern (DatabaseName|Schema).meta.xml
+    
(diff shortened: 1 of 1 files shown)


docs/source/configuration/commandline.rst
```diff
@@ -82,8 +82,10 @@ Processing
 Additional data
 ~~~~~~~~~~~~~~~
 [-meta fileOrPath]
-    Single schema analysis file path to SchemaMeta-xml, when running ``-all`` or ``-schemas`` path to directory containing
-    SchemaMeta-xmls with pattern (DatabaseName|Schema).meta.xml
+    Single schema analysis file path to one (1) SchemaMeta-xml-file;
+    when running ``-all`` or ``-schemas`` it defines the path to directory containing one ore more SchemaMeta-xml-files with pattern (DatabaseName|Schema).meta.xml
```
npetzall, line 86: ore should probably or.

When analyzing single schema, should be path to a single schemameta <- (could insert link here)
When analyzing multiple schemas (``-all`` or ``-schemas``), should be path to a folder containing schemameta files with naming pattern [databaseName or schemaName].meta.xml I.e. You have schemas backoffice and frontoffice in your database to augment these with schemameta you would need to specify a folder as ``-meta`` and that folder should contain backoffice.meta.xml and frontoffice.meta.xml

aisbergde: I used trial and error and after getting some errors.
That's why I think it could be explained better to get less errors.

Finally this worked fine and I added this example:

```
# one schema
schemaspy.s=config
# one path
schemaspy.meta=SchemaMeta
```

and I think this will work also with `-schemas` and `-all`

You can modify my request and improve it :-)

npetzall: Actually I think it's a weird implementation to allow file or folder. It would be much simpler to just say folder and it should contain file(s) named [DatabaseName or SchemaName].meta.xml, just allow one single use case.

Hmmm, even if it supports specifying a file, we could just document the folder usecase.....

aisbergde: I agree, and if there is only one file it can also be in the folder. And if someone uses the file, then this will be a hidden feature.
Anyway may be this will change a little bit with 7.0 if the meta definition will be extended?

aisbergde:
you could do.
and you could create a separate faq sub page for ms sql server so we could post database specific content there

npetzall:
@aisbergde I've started with the FAQ for specific database. But I'm a bit unsure if it should be under FAQ it will become very nested. But what do you think about:
bild


Source code layout 
bild

Database\index.rst only has a header and will then get includes for each database that we want documentation for.
So currently it would look like:
```
.. |br| raw:: html

   <br />

*********
Databases
*********

.. include:: mssql.rst.inc
```

And the mssql.rst.inc would just start with
```
Microsoft SQL Server
====================

.. _singlesignon:

Single Sign-On
--------------

mssql
^^^^^
When running SchemaSpy |br|
**before** ``-jar SchemaSpy-[version].jar`` |br|
add ``-Djava.library.path=[path-to-dir-containing-sqljdbc_auth.dll]`` |br|
**after** ``-jar SchemaSpy-[version].jar`` |br|
add ``-sso`` |br|
When using cmd add ``-connprops integratedSecurity\=true`` |br|
When using git bash in windows add ``-connprops integratedSecurity\\=true`` |br|

mssql-jtds
^^^^^^^^^^
When running SchemaSpy |br|
**before** ``-jar SchemaSpy-[version].jar`` |br|
add ``-Djava.library.path=[path-to-dir-containing-ntlmauth.dll]`` |br|
**after** ``-jar SchemaSpy-[version].jar`` |br|
add ``-sso`` |br|
```

aisbergde:
I think this is a good way

PR #710, opened 2020-06-19
Add Snap distribution

edumucelli:
Congratulations for the work done on Schemaspy, it is great!

In order to easy Schemaspy installation, this PR adds the snap distribution. For [Snapcraft](https://snapcraft.io) users it will just require a `snap install schemaspy` command and they would be good to go.

However that still requires the `Schemaspy` snap being available in the [Snapcraft store](https://snapcraft.io/store). I can do that as well in case are willing to accept this PR.


edumucelli:
Hi @npetzall, hope you are feeling better. Any news on that one? Cheers!

npetzall:
Illness turned into vacation. I've had a look and as I understand it this is just a snap instructions. Still needs to be added to snapstore?

npetzall:
I've read up a bit on snapcraft. We probably need an brand/umbrella. I'll create a personal account. Since @rafalkasa is the owner of the github I feel that he should create the snapcraft brand/umbrella account.

@rafalkasa what do you say do you have time to set up a brand/umbrella account at snapcraft.io?

edumucelli:
Yes, correct, those are the instructions to generate the Snap. It can be installed as it is locally, but it won't be available for general public. Go ahead and create the account, better you guys on the upstream have the account. You can try the `snapcraft` proposed here by:

```
cd snap
snapcraft
snap install schemaspy_6.1.0_amd64.snap --dangerous
```

The `dangerous` bit is just because it is not a snap that comes from the Snap store yet, it was built locally so it is not signed. Once you have done that you can use `schemaspy` executable directly (without the `java jar`).

Issue #489, opened 2019-01-02
PostgreSQL v11 - inconsistent/incorrect column "Size"


zam6ak:
This was briefly mentioned in #395, but I am breaking it out into this separate issue.

On the table detail view, in the Columns grid, "Size" is incorrectly reported for various column types.
This is especially true for numeric types which distinguish between "Size" (as in storage size in bytes) vs "Range"
https://www.postgresql.org/docs/current/datatype-numeric.html

For example, take a look at the table `public.film` from sample DB (details below)

* column `title` is `varchar(255)` - Size shows 255  (I guess this is ok because column is defined as `varchar(255)`)
* column `description` is `text` - Size shows 2147483647 (???)
* column `film_id` is `serial` - Size shows `10` (10 what? serial columns in PG are actually `integer` columns of 4 bytes)

and so on...

So the first question is, what is Size supposed to represent?
a) storage size (in bytes) for given type
b) precision in case of decimal numerical types and relevant size of text types
c) something else

in either case, the information seems mixed and incorrect

---
Tested using:

* DB:        PostgreSQL v11
* Driver:    [PostgreSQL JDBC 4.2 Driver, 42.2.5](https://jdbc.postgresql.org/download/postgresql-42.2.5.jar)
* Schemaspy: [schemaspy-6.1.0-SNAPSHOT](https://github.com/schemaspy/schemaspy/raw/gh-pages/schemaspy-6.1.0-SNAPSHOT.jar)
* Sample DB: [DVD Rental](http://www.postgresqltutorial.com/postgresql-sample-database/)
* Java ver:  1.8

npetzall:
I looked at the code and it seems to be "it depends" on what it represents. So we should make it more clear.

Varchar it is number of characters
For string is probably maximum length. I probably relates to how long a string can be in Java. But number of characters.
For serial is actually precision of the integer how many digits that can be stored including decimals. Serial can store 1 to 2147483647 where the latter is 10 digits. 

Lenght might be a better name for it since it's more of a Lenght thing than a size. Might need to Google and see if there exists a more appropriate term.

zam6ak:
I think "it depends" approach is going to confuse a lot of people (not just me)
FWIW, displaying storage size (how many bytes type takes) or range of particular type (10 digits, or max 32 bit integer (i.e. 2^31 -1) is just not "information worthy", IMHO....Each DB manual will have that information stated for particular types....


However, precision for numeric types or length of varchar column are very useful and should show.
[...]

PR #744, opened 2021-01-22
Removed (Number)rs.getObject(BUFFER_LENGTH) from calculation of colum…

lelle1234:
…n length

### DIFF
--- a/src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
+++ b/src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
@@ -89,14 +89,8 @@ public class ColumnService {
         tmp = rs.getString("TYPE_NAME");
         column.setTypeName(tmp == null ? "unknown" : tmp.intern());
         column.setType(rs.getInt("DATA_TYPE"));
-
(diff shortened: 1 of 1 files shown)


npetzall:
I checked the tests and saw which are failing. Only one actually or "one set". Which i MSSQL, the int column went from 10 to 4.
I looked into it, when I figured out the 10, I realized I had done the same thing previously back in #489.

So I wonder if the most "correct" way might be to actually store both size and buffer length.
Rename size column to "Allocation" or "Space" and have it populated with the buffer_length.
Add column size and precision to the actual type information so 
Type: int[10]
Allocation: 4

Or as in #489 just remove the size column. Since datatype allocation can be checked in manual.

lelle1234:
I think you are right, these are two different aspects of the type. The
logical size related to DDL, and the physical size related to storage.

The logical size is irrelevant for certain types such as integer, but I
don't think it is worth while to introduce an exception for these. Logical
size is likely the more interesting of the two for most people.

Den sön 24 jan. 2021 00:20Nils Petzäll <notifications@github.com> skrev:

PR #748, opened 2021-03-11
Update the Install Requirements, adding JDBC driver

### LINKED ISSUE #747: Docs: Document where to Download Missing JDBC Drivers, such as for Postgres
## Expected Behavior
I assumed the latest drivers would be pre-installed.

## Current Behavior
When I ran the one-liner on a Postgres connection it returned:

> There were missing paths in driverPath:
> 	/org/schemaspy/drivers/postgresql-9.4.1208.jre6.jar
> 

However, I was able to resolve this by manually tracking down [a proper driver](https://jdbc.postgresql.org/download.html) and using the -dp flag.

## Possible Solution
Was my experience an edge case, or is this likely to happen to anybody using Postgres?

Or, is there's already documentation on this that I missed? If so, I feel it could be better promoted because I looked through the entire [Docs](https://schemaspy.readthedocs.io/en/latest/index.html) and this Github repo directories and ended up having to figure it out myself based on prior experience.

In any case, I'll offer to help add basic document about how to locate a proper JDBC drivers in the event there's a failure.

For example, the latest [JDBC Postgres driver](https://jdbc.postgresql.org/download.html) worked by adding the flag `-dp postgresql-42.2.19.jar`

## Context
When I received an error I turned to the docs, but didn't see anything about how to source the correct driver.

I then assumed the drivers could be found in the Github repo, but didn't locate them.

I decided to do an online search for "JDBC Postgres" because I'm familiar with ODBC concepts from work in the early 2000's with MS Access. Luckily, I landed on [the page](https://jdbc.postgresql.org/download.html) that happen to have the file I needed.

However, I don't know if many new web developers know about JDBC and would quickly figure out that they needed to manually find and download a JDBC driver.

allella:
Here's a first crack at adding the JDBC driver requirement to the documentation and updating the requirements section.

I don't know if you officially support OpenJDK, but I added that as an option since it's working on my system.

Also, I tried to run `make make.bat` and `sphinx-build`, but they failed.  I saw a note in one of the other issues that Readthedocs probably does the compile step. I did include a few internal :ref links, but I can't really test them without something compiling the files.

Closes #747 

- Add a list of links to likely JDBC download locations for most of the supported engines
- Converted external links to text links
- Updated the Java section to include OpenJDK as an option
- Rework the viz.js or Graphviz requirement section

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -11,7 +11,7 @@ Before you can use SchemaSpy you must have the following requirements available 
 * viz.js or Graphviz
 
 Java
-^^^^^^
+^^^^
 
 You will need to have a supported version of Java installed, which includes:
 
@@ -24,7 +24,8 @@ If you don't already have a proper version of Java installed, see `OpenJDK <http
 
 
 JDBC Driver
-^^^^^^
+^^^^^^^^^^^
+
 No JDBC drivers are included with the jar-distribution of SchemaSpy.
 
 In some case, a JDBC driver may already exist on your local system if your database happens to come with drivers. Otherwise, you will need to download one yourself.
@@ -53,15 +54,15 @@ You will need to tell SchemaSpy where to find the JDBC driver you require. For e
 
 If your JDBC .jar file is in a different directory, then the -dp parameter would need to specify the directory path to the file, like ``-dp /opt/some-directory/your-jdbc-driver-name.jar``. 
 
-See :ref:`_commandline` for more information and advanced cases, such if you need to load multiple driver files with ``-loadjars``.
+See :ref:`commandline` for more information and advanced situations.
 
 viz.js or Graphviz
-^^^^^^^^
+^^^^^^^^^^^^^^^^^^
 This is necessary to render graphical representations / images of the database relationships.
 
 SchemaSpy version 6.1.0 and higher now comes with viz.js, so you will not need to download anything unless you're using SchemaSpy version 6.0 or lower.
 
-For SchemaSpy version 6.1.0 or higher, simply include the ``-vizjs``  :ref:`_commandline` when executing the SchemaSpy command.
+For SchemaSpy version 6.1.0 or higher, simply include ``-vizjs`` as a command line argument when executing the SchemaSpy command.
 
 If you must use Schemaspy version 6.0 or less, then Graphviz will need to be installed as follows.
 
@@ -79,7 +80,7 @@ If you must use Schemaspy version 6.0 or less, then Graphviz will need to be ins
     Please read carefully the detailed instructions on how to `install Graphviz on your operating system <http://www.graphviz.org/download/>`_.
 
 SchemaSpy
-----------
+---------
 
 SchemaSpy is just a single executable jar-file (schemaspy-[version].jar), you can `download releases from the SchemaSpy website <http://schemaspy.org>`_ or the `GitHub releases page <https://github.com/schemaspy/schemaspy/releases>`_.
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -4,20 +4,66 @@ Installation
 Requirements
 ------------
 
-Before you can start using SchemaSpy you must have installed two things in your system environment.
+Before you can use SchemaSpy you must have the following requirements available on your local system.
 
-Java 8
+* Java
+* a JDBC driver file for your specific database engine 
+* viz.js or Graphviz
+
+Java
 ^^^^^^
 
-Download instructions for all operating systems: `https://java.com/en/download/manual.jsp <https://java.com/en/download/manual.jsp>`_
+You will need to have a supported version of Java installed, which includes:
 
-Optional
---------
+* Java version 8 or higher
+* OR OpenJDK 1.8 or higher - an open-source alternative
 
-Since version 6.1.0 Graphviz is no longer required. There is an embedded viz.js that can be used by adding command line argument ``vizjs``
+You can run ``java -version`` in a terminal to check the version of any currently installed Java.
 
-Graphviz
+If you don't already have a proper version of Java installed, see `OpenJDK <https://openjdk.java.net/install/>`_ or `Oracle Java <https://www.oracle.com/java/technologies/javase-downloads.html>`_ for download and install instructions for your operating systems.
+
+
+JDBC Driver
+^^^^^^
+No JDBC drivers are included with the jar-distribution of SchemaSpy.
+
+In some case, a JDBC driver may already exist on your local system if your database happens to come with drivers. Otherwise, you will need to download one yourself.
+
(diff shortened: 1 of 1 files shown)


allella:
The changes are easiest to see in GitHub at https://github.com/allella/schemaspy/blob/docs/update-install-requirements/docs/source/installation.rst

allella:
@npetzall who would be the appropriate person to review a docs PR?
Thanks

npetzall:
Hi, sorry been a bit busy I'll try to do it shortly.

docs/source/installation.rst
```diff
@@ -4,20 +4,66 @@ Installation
...
+* Java version 8 or higher
+* OR OpenJDK 1.8 or higher - an open-source alternative
+
+You can run ``java -version`` in a terminal to check the version of any currently installed Java.
+
+If you don't already have a proper version of Java installed, see `OpenJDK <https://openjdk.java.net/install/>`_ or `Oracle Java <https://www.oracle.com/java/technologies/javase-downloads.html>`_ for download and install instructions for your operating systems.
+
+
+JDBC Driver
 ^^^^^^
```
npetzall, line 27: Title underline too short.

docs/source/installation.rst
```diff
@@ -4,20 +4,66 @@ Installation
...
+
+See :ref:`_commandline` for more information and advanced cases, such if you need to load multiple driver files with ``-loadjars``.
+
+viz.js or Graphviz
 ^^^^^^^^
+This is necessary to render graphical representations / images of the database relationships.
+
+SchemaSpy version 6.1.0 and higher now comes with viz.js, so you will not need to download anything unless you're using SchemaSpy version 6.0 or lower.
+
+For SchemaSpy version 6.1.0 or higher, simply include the ``-vizjs``  :ref:`_commandline` when executing the SchemaSpy command.
```
npetzall, line 64:
```
:ref:`_commandline`  

 should be without the leading "_"

File: docs/source/installation.rst [hunk 3]
```diff
@@ -4,20 +4,66 @@ Installation
...
+* `SQLite <https://github.com/xerial/sqlite-jdbc>`_
+* `SQL Server <https://docs.microsoft.com/en-us/sql/connect/jdbc/download-microsoft-jdbc-driver-for-sql-server>`_
+* `Sybase <http://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.help.sqlanywhere.12.0.1/dbprogramming/jconnect-using-jdbxextra.html>`_
+* `Teradata <https://downloads.teradata.com/download/connectivity/jdbc-driver>`_
+
+You will need to tell SchemaSpy where to find the JDBC driver you require. For example, if you downloaded the Postgres JDBC file called postgresql-42.2.19.jar to the current directory the command would include the -dp parameter, like ``-dp postgresql-42.2.19.jar``. 
+
+If your JDBC .jar file is in a different directory, then the -dp parameter would need to specify the directory path to the file, like ``-dp /opt/some-directory/your-jdbc-driver-name.jar``. 
+
+See :ref:`_commandline` for more information and advanced cases, such if you need to load multiple driver files with ``-loadjars``.
```
npetzall, line 56:
```
:ref:`_commandline`  

 should be without the leading "_"
```

docs/source/installation.rst
npetzall: The "built" line will be:
`For SchemaSpy version 6.1.0 or higher, simply include the -vizjs Command-Line Arguments when executing the SchemaSpy command.`   
I wonder if the ref should be removed in this case.

docs/source/installation.rst
npetzall: Documentation needs to be updated in regards to `-dp` it will traverse the path and pick upp alla folders and jars.
So the `-loadjars` is sort of deprecated. There's a single nail left in the coffin for it. It's best to specify the file but for drivers with multiple jars. It should be possible to just specify the folder which contains the jars.

docs/source/installation.rst
allella: That's fine. I noticed the use of "arguments" and "parameters" are not consistent.  Is there a preference?  Also, would you want the other docs to be consistent? 

"Parameters" seems to be the less commonly used word in the context of a shell script. I personally would associate "parameters" with functions and classes inside a script and "arguments" coming in from outside the script.

docs/source/installation.rst
allella: Increased the underlining in the latest commit.

docs/source/installation.rst
allella: Removed the _ in the ref and changed the -vizjs to say "command line argument" instead of a ref.

docs/source/installation.rst
allella: - Removed the _ in the ref.
- Removed the part about -loadjars.

npetzall (approved):

allella:
Thanks for merging this in.  @npetzall is there a manual step or automated schedule on the ReadTheDocs updates?  These changes [aren't reflected on the docs page](https://schemaspy.readthedocs.io/en/latest/installation.html#requirements).

npetzall:
Should be built automatically, but it will end up I latest/snapshot. I'll have a look

allella:
I see this made it to [master](https://schemaspy.readthedocs.io/en/master/installation.html
), but not latest or v6.1.0, so it's not showing for new visitors to the docs.  v6.1.0 seems to be the default if I visit https://schemaspy.readthedocs.io in a new browser session.

Best,
Jim

npetzall:
You're right. 

We changed that a while back due to unreleased features appearing in the docs under latest. So we switched to latest release.

Since we don't have release branches, we can't update the documentation in the past. So it will eventually become the latest.

But we should probably consider release branches since we don't do releases that often.

allella:
Yeah, I suppose that's all up to you, but I think it's worth getting the documentation changes out there.  Of course I'm biased, but I think changes I made should clear up confusion in a couple areas I had as a newbie trying to figure out the install steps.

Semantic versioning is so common these days I don't see it confusing people as long as no breaking changes end up in a [patch release](https://semver.org/).

I did notice some inconsistencies in the other parts of the docs, like the use of "arguments" vs "parameters" vs "flags" to refer to the Java binary's command line arguments.  I could take a crack at a PR to clean things up a bit if you'd like the release to have more substantial changes in the docs.

npetzall:
We could do patch releases, we actually had the idea of doing releases every 3 months. But that was back when we had the time.

We can still create release branches from existing tags, but the burden of maintaining multiple release would be to big. So push forward with the 3 month release cycle or if it was every month. We ended up with nothing.

I have some issues that needs to be fixed with the build, and there might be a person coming onboard to help with development. I'm gonna say that person is undecided as of now.

It would be awesome if you took a crack at that. I'll try to free up some time and I'll try to fix the build issues and start doing releases. It would be a nice step forward after hmm, almost 2 years (yeah I rounded up to make it sound more important).

PR #756, opened 2021-04-10
Add warning about Nashorn in Java 15

ugultopu:
I also believe a build of the docs is required, since [the docs](https://schemaspy.readthedocs.io/en/latest/installation.html) do not have the latest version.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -64,7 +64,7 @@ SchemaSpy version 6.1.0 and higher now comes with viz.js, so you will not need t
 
 For SchemaSpy version 6.1.0 or higher, simply include ``-vizjs`` as a command line argument when executing the SchemaSpy command.
 
-NOTE: If you are using Java 15 or later, ``vis.js`` will not work! This is because Nashorn (JavaScript engine) was removed on Java 15. Hence, if you are on Java 15 or later, you need to install Graphviz manually as described below. Or you can switch to an earlier Java version if you would like to use ``vis.js``.
+NOTE: If you are using Java 15 or later, ``vis.js`` will not work! This is because Nashorn (JavaScript engine) was removed on Java 15. Hence, if you are on Java 15 or later, you need to install Graphviz manually as described below. Or, you can switch to an earlier Java version if you would like to use ``vis.js``.
 
 If you must use Schemaspy version 6.0 or less, then Graphviz will need to be installed as follows.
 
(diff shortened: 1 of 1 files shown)


npetzall:
Our build is currently broken and I'm m working on that. We made a strategy way back that we should release more often and skip release branches. So there is some issues with docs that needs to be figured out.

But we could implement a simple check and fail fast if vizjs is selected and a Javascript engine is missing.

ugultopu:
Is there anything I can help with creating a new release? Which issues need to be fixed before a new release can be created?

Also, do you think this PR is worth merging? I believe it is because adding this warning to the README might help users to understand why `viz.js` isn't working and this could prevent opening of duplicate issues in the issue tracker.

npetzall:
I was thinking of a patch release as soon as I get the build working with github actions. Since Travis has become a bit unreliable in regards to docker.

There are other benefits when maintainers are added one less system to care about.

I'm almost done so a patch release should be doable in a few days.

The there's the docs thingy. That I think might have stopped working so need to check that. But latest points to latest release. Since we have no release branches we can't update docs for past releases. But there is a master version which contains the latest docs changes in master and is related to the development version.

npetzall:
I think it's a good addition to the docs, but better error message for now would be good too. So if you want to take a stab at better error message that would be appreciated.

Or give a proposal of how we could  avoid the problem. Currently I'm thinking maybe bloat the jar by adding a script engine or have two different binaries with and without a script engine.

But our normal recommendation would be to use docker. Hmmm, I wonder if that's in the docs???

ugultopu:
> I think it's a good addition to the docs, but better error message for now would be good too.

That makes sense. IMHO:
- As the first step, we can merge this PR so that a user that reads the documentation from https://schemaspy.readthedocs.io/en/master/installation.html will be aware of the potential issue with `viz.js` and Java 15+.
- The second step would be, as you said, adding an error message when the user selects `vis.js` on a Java version 15+.
- The third and the real solution would be to fix the problem altogether by either embedding a JavaScript engine to SchemaSpy, or thinking of another solution.

> But our normal recommendation would be to use docker. Hmmm, I wonder if that's in the docs???

That sounds like a much better idea too. I'm going to look at the docs and create another PR if it isn't.

npetzall:
I've fixed the docs integration so, the updates should be available in the master "version".

npetzall:
And the docs, doesn't even mention docker =(

ugultopu:
> And the docs, doesn't even mention docker =(

Right, it definitely should be in the docs, since it is the most straightforward way of using SchemaSpy. I will try to create a PR for it but if you have availability to do so, that would be great as well 😊

npetzall:
I'll add docker to the docs. I'm just finishing up some build issues first, then I'll do the docs and after that I'll add the release workflow.

PR #762, opened 2021-04-18
chore(release): Github Actions Workflow

npetzall:


rafalkasa:
@npetzall this is looking awesome :-) we can merge it

npetzall:
Since this was back in April and the documentation for maven release isn't crystal clear I'll setup a simple repo and test it out.

npetzall:
Simple repo for testing -> https://github.com/npetzall/gha-maven-release-dev

Things that I've learned. 
POM might need updating to HTTPS in SCM.
Run for setting up git is needed: `git config user.name` `git config user.email`
The release-ready action needs updating

If you have a look in the simple repo. prepare is an understatement. It creates the tag, it builds the binary, it creates the next version. So it actually does a release in SCM. Which is good. Since we are almost done with this. I'll update it later. Other things to do right now.

npetzall:
I've updated the PR so that the release plugin. It will be a manually triggered action.

Possible enhancements would be to verify the version numbers better. Make sure there aren't any tags with current or next version already in the repository.

PR #768, opened 2021-05-04
Using estimate instead of count for pgsql

thiagoazeredo-quintoandar:
Change row counting on postgres to use estimate and reduce the load on database


rafalkasa:
@thiagoazeredo-quintoandar thank you for your contribution but I decided to merge PR #784 
I took decision based on post https://stackoverflow.com/questions/7943233/fast-way-to-discover-the-row-count-of-a-table-in-postgresql

PR #774, opened 2021-06-13
refactor(dot): Microtype name

jesperolsson-se:


src/main/java/org/schemaspy/output/dot/schemaspy/name/Null.java
```diff
@@ -0,0 +1,16 @@
+package org.schemaspy.output.dot.schemaspy.name;
+
+/**
+ * Null object for name.
+ */
+public final class Null implements Name {
```
npetzall, line 6: Null is a loaded word, and it's actually an EmptyName. Could we rename it to EmptyName?

npetzall (changes requested):

src/main/java/org/schemaspy/output/dot/schemaspy/name/Null.java
jesperolsson-se: Would `Empty` suffice ("name" can be inferred from the scope of the package and/or interface)?

npetzall:
```
Name diagramName = new Degree(            
  twoDegreesOfSeparation,            
  new Implied(               
     includeImplied,                
     new Empty()           
  )        
);
```

Compilers, IDEs are better at inferring than humans. Empty is generic term so I think it hurts readability/understandability.

Import statement is at the top of the file.

src/main/java/org/schemaspy/output/dot/schemaspy/DotTableFormatter.java
```diff
@@ -80,8 +84,15 @@ public DotTableFormatter(
 
         DotConnectorFinder finder = new DotConnectorFinder();
 
-        String diagramName = (twoDegreesOfSeparation ? "twoDegreesRelationshipsDiagram" : "oneDegreeRelationshipsDiagram") + (includeImplied ? "Implied" : "");
-        dotFormat.writeHeader(diagramName, true, dot);
+        Name diagramName = new Degree(
+            twoDegreesOfSeparation,
+            new Implied(
+                includeImplied,
+                new Simple("")
```
npetzall, line 91: Replace with Empty Name?

npetzall (approved):

PR #775, opened 2021-06-17
refactor(dot): Extract method for writing immediate relatives

jesperolsson-se:


src/main/java/org/schemaspy/output/dot/schemaspy/DotTableFormatter.java
```diff
@@ -243,4 +237,21 @@ private static void markExcludedColumns(Map<Table, DotNode> nodes, Set<TableColu
             }
         }
     }
+
+    private void writeImmediateRelatives(
+        Set<Table> relatedTables,
```
npetzall, line 242: What are your thought on returning a map and skipping having said map as an argument?

jesperolsson-se: I'd like to move in that general direction, but we're not really there yet. The main problem is that this piece of code has several side-effects: nodes and connectors (perhaps we're missing some sort of `Graph` abstraction?). I'd argue that eliminating just one of them would **decrease** understandability.

npetzall: I glanced over the code and for some reason I'm stuck up on temporal coupling.
So first I think 
```
        // connect the edges that go directly to the target table
        // so they go to the target table's type column instead
        for (DotConnector connector : connectors) {
            if (connector.pointsTo(table))
                connector.connectToParentDetails();
        }
```

Should also be moved to writeImmediateRelatives, it might increase the complexity, but I think the code belongs together for the "writeImmediateRelatives" concept (even if it's incorrect abstraction).

Returning the nodes map from writeImmediateRelatives forces the method to be called before the Map is used, it sort of enforces the comment.

jesperolsson-se: Yes, I agree that both are part of the same concept. Will update. (This is the primary reason I'm extracting the method: It clarifies scope better than comments.) 

I also agree with your arguments concerning temporal coupling. However, (again) I don't think we're there yet. The nodes and connectors are **both** used in later parts of the original method.

npetzall: Hmm, with the current implementation `writeImmediateRelatives` can be moved to line 179 without changing scope. 
If `writeImmediateRelatives` created and returned the Map `nodes` without changing scope it could only be moved to line 110.

Then again `// Immediate relatives should be written first` might be a mental model thing and not a requirement for dot-lang.

jesperolsson-se: I'm not certain I'm making myself understood, so I'll try again.

There are many components in understandability. Temporal coupling is one of them and should be addressed, as you say. However, I think doing so _now_ would be premature.

Another component is [command-query separation](https://martinfowler.com/bliki/CommandQuerySeparation.html). As long as this method writes _nodes and connectors_ but can return only one of them, the principle would be violated.

Another is that the choice would be (conceptually) arbitrary. Why should the method create and return nodes instead of creating and returning the connectors? Hence, we would introduce further confusion.

Each component can be addressed in several ways, but the real problem is that the file is very/needlessly complex. After some additional refactorings, I'm confident that we can resolve the temporal coupling _without_ compromising other components (too much).

PR #776, opened 2021-06-29
refactor(dot): Reduce method complexity

jesperolsson-se:


jesperolsson-se:
I'm noticing that some tests are failing now, even though everything passed locally when I followed the instructions for contribution. Are the tests stricter for PRs? I think it would be a welcome improvement if the instructions referenced how to execute the entire suite locally. For example, it would help me locate when I introduced the fault.

npetzall:
Hmm, I'm sure I have written that somewhere but git history doesn't lie so I must have written it somewhere else.

It's the standard verify goal. However it uses assumptions for docker and some jdbc drivers. So the github actions might test more than locally.

npetzall:
Summer vacation is almost over, so I'm home a lot more. Could you rebase of master. I've merged so that the output from integration tests are added as a workflow artifact. 

So that files can be diffed locally also.

Issue #550, opened 2019-04-24
Remove Spring


npetzall:
Usage of spring is limited and might be unnecessary so we should look into removing it.

At least weigh the effort for removal and pros/cons of doing so.

I'm personally for removal, the usage is small.
We could simplify some code when working with "resources".
We would need to implement wiring and also if we really like the startup + banner that needs to be copied/implemented.

PR #783, opened 2021-07-18
Drop Spring Boot

onkobu:
- remove it from pom.xml, import, deps, plugins
- bump versions, esp. Mockito and jcommander
- introduce logback explicitely
- introduce fixture for services, expose base services
- other references local variables
- fix resource loading through classloader
- fix resource processing
- fix command line parser test


onkobu:
Inspired by #550 and a sort of donation because using schema spy saved effort for me in the past. It doesn't fully compile yet but (src/test) but is a starting point. Main features work: resource loading, argument parsing and service creation.

The factory for default properties is a bit hacky but the testing hatch is locally enough. I didn't manage to get the integration tests working, lacking the Docker-runtime.

onkobu:
org.schemaspy.input.dbms.service.DatabaseServiceIT makes a design problem visible. Class Config and CommandLineArguments co-exist. There are some deprecations in Config suggesting CommandLineArguments are more recent. The test itself creates useless mocking. The services don't use any of the CommandLineArguments properties.

npetzall:
The idea is to transition to commandline args. Since Config has a lot of issues. But that work has been underway for a cpl of years.

onkobu:
> transition to commandline args

I'd say this is to be set apart from dropping Spring. Nevertheless doable in finite time. The parser digests args array and fills the arguments entity – as simple as deleting Config and replacing all occurences with arguments entity. This feels like an application context/ general state record.

Now the PR compiles and surefire runs fine. Failsafe hiccups with DotConfig-class – NoClassDefFoundError. Doesn't look like a JDK issue (same on 8 and 11).

And there'll be plenty of work left for migration to JDK 11. I won't do this neither to limit the change set.

onkobu:
With the last commit also Failsafe runs fine. Dropping Spring from POM returns to a plain JAR thus no manipulation necessary to get test classpath based on JAR back working. (Spring Boot JARs aren't common JARs with their bootstrapping stuff and library ingestion.)

May I finally suggest to consolidate integration test filtering? There's <exclude> in pom.xml as well as `@Ignore` in some of the classes. Maybe even revive the ignored ones or simply drop them.

npetzall:
It wasn't meant as a task (command line args), only an explanation to the state. Just doing the replacement would still require static instances which is in my opinion the biggest issue with Config.

I'll have a look at this hopefully during the week.

As for the tests, the failsafe excludes are related to the inclusion of *Suite.class, can be looked over but, i think I had the issue of tests running twice both from the suite and then as they were named IT. Could be solved with different naming convention.

I actually made changes to the structure and output of tests just prior to the weekend and most of the `@Ignored` should removed. I think there is a MariaDB test that might be interesting to keep as ignored just-in-case any issues are reported. But since it a drop-in replacement for MySQL there shouldn't be any issues.

> org.schemaspy.input.dbms.service.DatabaseServiceIT makes a design problem visible. Class Config and CommandLineArguments co-exist. There are some deprecations in Config suggesting CommandLineArguments is more recent. The test itself creates useless mocking. The services don't use any of the CommandLineArguments properties.

God catch, the test hasn't been maintained. CommandLineArguments needed to exist in the SpringContext because it was injected into SqlService, TableService it isn't anymore.

It's built with JDK 11, since Sonar doesn't support JDK 8. But moving to JDK 11 or even 16 with included minimal JRE we need to implement JDK modules(jigsaw).

PR #784, opened 2021-07-23
Approximate row count for postgres

unusual-thoughts:
As mentioned [here](https://wiki.postgresql.org/wiki/Slow_Counting), row count can be extremely slow for postgres with large tables.
This introduces custom SQL as recommended [here](https://wiki.postgresql.org/wiki/Count_estimate) for postgres instead of `count(*)` to get approximate row counts instead of hanging indefinitely for huge tables


thiagoazeredo-quintoandar:
I've open a PR to fix it too: https://github.com/schemaspy/schemaspy/pull/768

src/main/resources/org/schemaspy/types/pgsql.properties
```diff
@@ -52,3 +52,5 @@ order by p.specific_name, p.ordinal_position
 selectCheckConstraintsSql=select tc.table_name, tc.constraint_name, cc.check_clause as text from information_schema.check_constraints cc left join information_schema.table_constraints tc on tc.constraint_catalog = cc.constraint_catalog and tc.constraint_schema = cc.constraint_schema and tc.constraint_name = cc.constraint_name where cc.constraint_name not like '%not_null' and tc.constraint_schema = :schema
 
 selectSequencesSql=SELECT seqs.sequence_name, seqs.start_value, seqs.increment FROM information_schema.sequences seqs WHERE seqs.sequence_schema = :schema
+
+selectRowCountSql=SELECT reltuples AS row_count from pg_class WHERE oid = CAST(:schema || '.' || :table AS regclass)
```
rafalkasa, line 56: Based on the Stack Overflow explanation solution proposed by @unusual-thoughts will make Better estimate.
https://stackoverflow.com/questions/7943233/fast-way-to-discover-the-row-count-of-a-table-in-postgresql

rafalkasa (approved):
@unusual-thoughts Thank you for your contribution is very appreciated

PR #794, opened 2021-09-29
fix problem when the table name is not quoted and there is more than …

DanNYSPD:
## Description 
Fixes the  problem when the table name is not quoted and there is more than one table name with same characters but different case, also I add the tablename in the message of InconsistencyException for error report purposes

## Expected Behavior

When there are two tables (or more) with the same characters but different case it must be able to avoid confussion among all this tables. 
For example, in postgres the table user and "User" are different tables (is case sensitive) so is necessary to use quotes to avoid confusion when making queries in a linux/unix Server. When you query <select * from user> is not the same to <select * from "User"> in a linux environment. (note that I use angle brackets to simplify the use of quotes)

At least a more detailed message is expected.

## Current Behavior
When there are this two tables and InconsistencyException exception is thrown when a column is not found in the column of the table object. Besides no table name is provided in the current exception message.

src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
`throw new InconsistencyException("Column information from DatabaseMetaData differs from ResultSetMetaData, expected to find column named: '"+ columnName + "' in " + listColumns(table));`

`throw new InconsistencyException("Column information from DatabaseMetaData differs from ResultSetMetaData, expected to find column named: '"+ columnName + "' in columns" + listColumns(table) +" of table: "+tableName);` 

## Possible Solution
This PR adds more info in the exception message(return the table name) and fix the problem by setting the flag forceQuotes to true.

## Context

* Database type: Postgres 
*  Command is showed below:
~~~bash
java -jar schemaspy-6.1.0.jar -t pgsql -dp $DP -db $DATABASE -host $SERVER -port $PORT -s public -u $USER -p $PASSWORD -o $DIRECTORY
~~~

## Your Environment
* Fedora 34
* JAVA 11

### DIFF
--- a/src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
+++ b/src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
@@ -43,7 +43,7 @@ public class ColumnService {
     public void gatherColumns(Table table) throws SQLException {
         initColumns(table);
         if (!(table.isView() || table.isRemote())) {
-            initColumnAutoUpdate(table, false);
+            initColumnAutoUpdate(table, true);
         }
     }
 
@@ -129,15 +129,15 @@ public class ColumnService {
         // so we can ask if the columns are auto updated
         // Ugh!!!  Should have been in DatabaseMetaData instead!!!
         StringBuilder sql = new StringBuilder("select * from ");
-        sql.append(sqlService.getQualifiedTableName(
-                table.getCatalog(),
-                table.getSchema(),
-                table.getName(),
-                forceQuotes)
-        );
+        String tableName=sqlService.getQualifiedTableName(
+            table.getCatalog(),
+            table.getSchema(),
+            table.getName(),
+            forceQuotes);
+        sql.append(tableName);
 
         sql.append(" where 0 = 1");
-
+        //LOGGER.info("SQL:"+sql.toString());
         try (PreparedStatement stmt = sqlService.getDatabaseMetaData().getConnection().prepareStatement(sql.toString());
              ResultSet rs = stmt.executeQuery()) {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
```diff
@@ -129,15 +129,15 @@ private void initColumnAutoUpdate(Table table, boolean forceQuotes) {
         // so we can ask if the columns are auto updated
         // Ugh!!!  Should have been in DatabaseMetaData instead!!!
         StringBuilder sql = new StringBuilder("select * from ");
-        sql.append(sqlService.getQualifiedTableName(
-                table.getCatalog(),
-                table.getSchema(),
-                table.getName(),
-                forceQuotes)
-        );
+        String tableName=sqlService.getQualifiedTableName(
```
DanNYSPD, line 132: Store the qualified table name to use both in the sql and in the exception message

src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
```diff
@@ -146,7 +146,7 @@ private void initColumnAutoUpdate(Table table, boolean forceQuotes) {
                 String columnName = rsMeta.getColumnName(i);
                 TableColumn column = getColumn(table, columnName);
                 if (Objects.isNull(column)) {
-                    throw new InconsistencyException("Column information from DatabaseMetaData differs from ResultSetMetaData, expected to find column named: '"+ columnName + "' in " + listColumns(table));
+                    throw new InconsistencyException("Column information from DatabaseMetaData differs from ResultSetMetaData, expected to find column named: '"+ columnName + "' in columns" + listColumns(table) +" of table: "+tableName);
```
DanNYSPD, line 149: improved message detail

rafalkasa:
This PR is ok for me and can be merged if small line will be corrected

npetzall:
> This PR is ok for me and can be merged if small line will be corrected

What needs correcting?

rafalkasa:
> @rafalkasa
> 
> > This PR is ok for me and can be merged if small line will be corrected
> 
> What needs correcting?

I'm referring to the comment with not used code 

`//LOGGER.info("SQL:"+sql.`

src/main/java/org/schemaspy/input/dbms/service/ColumnService.java
```diff
@@ -129,15 +129,15 @@ private void initColumnAutoUpdate(Table table, boolean forceQuotes) {
...
+        String tableName=sqlService.getQualifiedTableName(
+            table.getCatalog(),
+            table.getSchema(),
+            table.getName(),
+            forceQuotes);
+        sql.append(tableName);
 
         sql.append(" where 0 = 1");
-
+        //LOGGER.info("SQL:"+sql.toString());
```
npetzall, line 140: Remove this comment

npetzall (changes requested):

npetzall (approved):

PR #795, opened 2021-09-30
Relations with column details

tammoippen:
I wanted to have the large relationship diagram with column types. I think, especially for smaller schema, this gives a really nice and full overview of the table structure. I wanted to give this a new 

I am not very familiar with the architecture and I am not sure if I broke something here (I could not get the tests running locally).

Also, there is a misspelled graph attribute in the dot files: 'ratio~n~' https://graphviz.org/docs/attrs/ratio/


npetzall:
@jesperolsson-se you've done work in this area, can you at a glance explain what change this is? 

I'm gonna sit down later and try to figure it out.

jesperolsson-se:
> @jesperolsson-se you've done work in this area, can you at a glance explain what change this is?

@npetzall Not really, unfortunately. If I interpret the original comment correctly, it could be as simple as not knowing about
`-maxdet` and the default value being too low.

PR #807, opened 2021-12-31
feature: DataTables Pagination options

vbro:
Add parameters to control paging, pageLength and lengthChange options for DataTables
See https://datatables.net/reference/option/


vbro:
Allows user to specify DataTables paging parameters, for example:
* ```-columnPageLength 200```
Adds html attribute `data-page-length=200` to `column_table` table in [column.html](https://github.com/schemaspy/schemaspy/blob/master/src/main/resources/layout/column.html)  
This sets the default page length for `column_table` table to 200


* ```-noDbObjectPaging```
Adds html attribute `data-paging=false` to `database_objects` table in [main.html](https://github.com/schemaspy/schemaspy/blob/master/src/main/resources/layout/main.html)
This displays all items in `database_objects` table in one page


* ```-indexLengthChange```
Adds html attribute `data-length-change=true` to `indexes_table` table in [table.html](https://github.com/schemaspy/schemaspy/blob/master/src/main/resources/layout/tables/table.html)
This allows the user of the generated webpage to choose a page length for `indexes_table` table

vbro:
CC: @npetzall and @rafalkasa for comment and review

npetzall:
@vbro I would like to avoid using Config.class since it's a mess, so using JCommander would be better.

I'm gonna look into inheritance in Mustache and what's possible in JCommander.

I feel that it would have been good to have a object that represent DataTableConfig that has three attributes that can be used with the inheritance in Mustache, alternative might be to have a function to create the config based on DataTableConfig.

But avoiding Config.class would be good, there is also a possibility of adding DataTableConfig inside HtmlConfig and have it added to the wrapping scope.

Motivation for HtmlConfig would be that it contains isPaginationEnabled() which might need to be divided for each table and not global.

Just noticed that I've forgotten the javascript config object that exists in the template which is another place to add the config.
But let me ponder it a bit. The function that we could create could create the complete config object.

vbro:
Another project idea is to deprecate [Config](https://github.com/schemaspy/schemaspy/blob/master/src/main/java/org/schemaspy/Config.java) and remove all references for clarity.

npetzall:
> Another project idea is to deprecate [Config](https://github.com/schemaspy/schemaspy/blob/master/src/main/java/org/schemaspy/Config.java) and remove all references for clarity.

We have a project for that, we should add depreciate and link to the github project. I'll create an issue for that.

vbro:
@npetzall @rafalkasa FYI new update to pull request

rafalkasa:
@vbro amazing work. @npetzall do you have any other things to the PR ?

Could you @vbro also extend documentation about this free new parameters [commandline.rst](https://github.com/schemaspy/schemaspy/blob/master/docs/source/configuration/commandline.rst)

npetzall:
@rafalkasa I'm gonna say that it's better to get it in than to have it perfect. Since perfect might become scope creep and it will never be finished.

Currently it's probably gonna fail the integration tests, since the output is changed, so actually getting some effort in to https://github.com/schemaspy/schemaspy/projects/5 would greatly improve the ability to contribute to the project without having to mess around with docker too much.

With that said, get it in and I can fix the tests afterwards.

vbro:
New parameters updated in [commandline.rst](https://github.com/schemaspy/schemaspy/blob/master/docs/source/configuration/commandline.rst)

rafalkasa:
I think that this PR can be merged now @npetzall what are you think ?

npetzall:
@rafalkasa go for it

rafalkasa:
@vbro thank you for your contribution

PR #810, opened 2022-01-11
Update FAQ instructions for MacOS

greyvugrin:


rafalkasa:
@npetzall sounds good for me :-)

npetzall:
@rafalkasa I see that wiki can only be edited by user with push access. This can be disabled and anyone can update the wiki. Should we put our trust in the github community....

Other options are FAQ.md and have pushes for that specific file that isn't "version" specific.

PR #827, opened 2022-02-13
chore(docs): sqlite instructions

npetzall:


npetzall:
It's old and moldy, so a quick review so that it's understandable. Should we include a link to an example db?

docs/source/databases/sqlite.rst.inc
```diff
@@ -0,0 +1,47 @@
...
+If we use chinook.db as an example we will have the following layout::
+
+    C:\Tmp\sqlite\schemaspy-6.1.1-SNAPSHOT.jar
+    C:\Tmp\sqlite\db\chinook.db
+    C:\Tmp\sqlite\driver\sqlite-jdbc-3.36.0.3.jar
+
+Place the command prompt in ``C:\Tmp\sqlite`` |br|
+Now you can run::
+
+    java -jar schemaspy-6.1.1-SNAPSHOT.jar -t sqlite-xerial -sso -o out -dp driver -db C:/Tmp/sqlite/db/chinook.db -s chinook -cat chinook
```
jesperolsson-se, line 41: Don't use snapshot.

docs/source/databases/sqlite.rst.inc
```diff
@@ -0,0 +1,47 @@
...
+
+Output will be written to ``C:\Tmp\sqlite\out`` |br|
```
jesperolsson-se, line 43: This is clear from the command

docs/source/databases/sqlite.rst.inc
jesperolsson-se, line 41: Split up up the command over several lines from readability.

jesperolsson-se (changes requested):

docs/source/databases/sqlite.rst.inc
```diff
@@ -0,0 +1,47 @@
...
+Arguments: |br|
+``-s`` same as database name
+``-cat`` same as database name
+``-db`` full path to database file
+
+Example layout on windows: |br|
+
+To avoid possible issues with space and strange characters. |br|
+Base directory ``C:\Tmp\sqlite`` |br|
+place ``schemaspy-[version].jar`` in the root of that folder. |br|
```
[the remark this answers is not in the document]
npetzall: Yes, is it a problem?

docs/source/databases/sqlite.rst.inc
```diff
@@ -0,0 +1,47 @@
...
+.. _instructions:
+
+Instructions
+--------------
+
+Recommended driver is the xerial driver: |xerial| |br|
+Driver can be downloaded from releases |br|
+To use the driver specify ``-t sqlite-xerial``
+
+If connecting without user/pass the ``-sso`` argument can be used.
```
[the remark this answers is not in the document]
jesperolsson-se: Are we gonna make that evaluation for each database that we support? I think it would be better to create a `credentials` page a refer to that one.

docs/source/databases/sqlite.rst.inc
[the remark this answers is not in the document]
jesperolsson-se: Yes, it's a maintainability problem. We don't want to repeat this for each database

docs/source/databases/sqlite.rst.inc
jesperolsson-se: In that case, do like we do in the `README` at tell them to browse `C:\Tmp\sqlite\out\index.html`. But, again, that's general instructions

Issue #466, opened 2018-11-25
Suggestion for the output of "relationships"


holgerjakobs:
The html file "relationships.html" shows an abridged list of attributes and calls it "compact", while the unabridged version is called "large".

The first problem is that it isn't obvious that these two words are clickable and switch between two different displays. There is just a blue line above one of them which is unintuitive.

The second problem is the naming "compact" and "large", which implies different font sizes or similar, but not displaying all or just some attributes.

## Expected Behavior
The buttons should be real buttons or otherwise marked as clickable.

The names should be "abridged" and "unabridged" instead of "compact" and "large".

## Current Behavior
see above

## Possible Solution
see above

## Steps to Reproduce (for bugs)
1. Open the relationship page.

## Context
not important

## Your Environment
not important

npetzall:
Look and feel can be changed, naming also.

Look and feel might be default Twitter bootstrap. Need to check, might be able to get it to feel mor as tabs or maybe flat buttons since it's been the style since windows 8 came. 

As for the abridge and unabridged. Would synonyms be okay. I have never heard the words before, I don't know if they are more common with in litrature than computing. Compact could be condensed maybe and Larg could be Full? Since large and compact refers to size. Diagrams are different in size but it's due to data loss and not smaller margins or smaller font and that should be communicated in the name.

npetzall:
To be consistent solution should applied to Tables diagram tabs.

PR #838, opened 2022-03-07
Refactoring

jesperolsson-se:


src/main/java/org/schemaspy/output/dot/schemaspy/name/DefaultName.java
```diff
@@ -0,0 +1,18 @@
+package org.schemaspy.output.dot.schemaspy.name;
+
+/**
+ * Decorates a name with the default name.
+ */
+public final class DefaultName implements Name {
```
npetzall, line 6: I wonder if it's easy within the domain to understand what a DefaultName is?

Default are generally tricky. Will default trigger if a value has been given?  

```
new Degree(
        twoDegreesOfSeparation,
        new DefaultName(
                new Implied(
                        includeImplied,
                        new EmptyName()
                )
        )
)
```

src/main/java/org/schemaspy/output/dot/schemaspy/name/Sized.java
```diff
@@ -0,0 +1,26 @@
+package org.schemaspy.output.dot.schemaspy.name;
+
+/**
+ * Decorates a name based on compactness.
+ */
+public final class Sized implements Name {
```
npetzall, line 6: Could you have a look at #466

src/main/java/org/schemaspy/output/dot/schemaspy/name/DefaultName.java
jesperolsson-se: I agree, "default" doesn't convey much information. Still, I found it to be more meaningful than e.g., `RelationshipsDiagram`. All DOT diagram names I've found in this part of the code base had this name (plus som extra bits details), so to me it seemed like some sort of default. If you can provide me with a class name than makes more sense, I'll be happy to change it!

Good point: `DefaultName` should probably not take any parameters. Instead, we could introduce something along `Concatenation` and/or `Priority` in [demonstrated here](https://github.com/jesperolsson-se/FizzBuzz-OCP-Challenge/tree/main/src/main/java/org/example/rearrange/positive).

src/main/java/org/schemaspy/output/dot/schemaspy/name/Sized.java
jesperolsson-se: Thank you for providing this link. Could you clarify "look at"? No decision seems to have been reached - do you want me to decide?

src/main/java/org/schemaspy/output/dot/schemaspy/name/DefaultName.java
npetzall: I'm in favor of mergin this and creating an issue for future improvement just so that it's not forgotten.

I don't trust myself in remembering. So I would like to create the issue first.

I would like your input on the following:

We remove the name RelationshipDiagram completely. It's redudent in the fact that we produce ERD.

We change the naming convention or post/pre-fix so that we could start with a BaseName. I'm gonna look into this a bit. It would strange to have default or base in the middle so either be in the beginning or end of the "nestling".

src/main/java/org/schemaspy/output/dot/schemaspy/name/Sized.java
npetzall: I should have been more clear. Decision would be great. But since you've spent time thinking about this. You might have additional input or thoughts.

This is not blocking there is a separate issue regarding naming of "size", abridged, unabridged.

src/main/java/org/schemaspy/output/dot/schemaspy/name/DefaultName.java
jesperolsson-se: > I would like to create the issue first.

Sure, no worries!

> We remove the name RelationshipDiagram completely. It's redudent in the fact that we produce ERD.

I agree, but don't really have a strong opinion.

> We change the naming convention or post/pre-fix so that we could start with a BaseName.

My stance is that pre-fix and post-fix (though I dislike the names) should be classes in their own right, more specifically [Composite](https://en.wikipedia.org/wiki/Composite_pattern)s. The base name and the "qualifiers" would be leafs (i.e., no longer decorators).

Issue #842, opened 2022-03-23
Dot/Name/DefaultName needs a better name


npetzall:
I wonder if it's easy within the domain to understand what a DefaultName is?

Default are generally tricky. Will default trigger if a value has been given?
```
new Degree(
        twoDegreesOfSeparation,
        new DefaultName(
                new Implied(
                        includeImplied,
                        new EmptyName()
                )
        )
)
```
Contributor Author
@[jesperolsson-se](https://github.com/jesperolsson-se) jesperolsson-se [11 days ago](https://github.com/schemaspy/schemaspy/pull/838#discussion_r825282264)

> I agree, "default" doesn't convey much information. Still, I found it to be more meaningful than e.g., RelationshipsDiagram. All DOT diagram names I've found in this part of the code base had this name (plus som extra bits details), so to me it seemed like some sort of default. If you can provide me with a class name than makes more sense, I'll be happy to change it!

Good point: DefaultName should probably not take any parameters. Instead, we could introduce something along Concatenation and/or Priority in [demonstrated here](https://github.com/jesperolsson-se/FizzBuzz-OCP-Challenge/tree/main/src/main/java/org/example/rearrange/positive).

Member
@[npetzall](https://github.com/npetzall) npetzall [11 days ago](https://github.com/schemaspy/schemaspy/pull/838#discussion_r825284485)

> I'm in favor of mergin this and creating an issue for future improvement just so that it's not forgotten.
> I don't trust myself in remembering. So I would like to create the issue first.
> I would like your input on the following:
> We remove the name RelationshipDiagram completely. It's redudent in the fact that we produce ERD.
> We change the naming convention or post/pre-fix so that we could start with a BaseName. I'm gonna look into this a bit. It would strange to have default or base in the middle so either be in the beginning or end of the "nestling".

Contributor Author
@[jesperolsson-se](https://github.com/jesperolsson-se) jesperolsson-se [11 days ago](https://github.com/schemaspy/schemaspy/pull/838#discussion_r825296638)

    I would like to create the issue first.

Sure, no worries!

    We remove the name RelationshipDiagram completely. It's redudent in the fact that we produce ERD.

I agree, but don't really have a strong opinion.

    We change the naming convention or post/pre-fix so that we could start with a BaseName.

[...]

PR #848, opened 2022-04-27
Refactoring

jesperolsson-se:


npetzall:
I would like to have the name connected to the processing. It would be interesting to see a spike if it's possible and which approach to name would work best.

I.e. The current name building would be instruction on how to build the graph and I would be able to build and get the name which would more or less just be a description of the instructions used. 

Within that name might need to change and that would be totally fine.

jesperolsson-se:
Is this blocking the PR? To me, this sounds like suggestions for future improvements. I find that changes are much smoother when not trying to refactor and change behaviour at the same time.

npetzall:
To be honest, I think you're playing it to safe.
This change is not needed, the idea is to have it removed. The name of the diagram has no true meaning.
It could be a guid as long as it's used in the correct place.(pointer/coordinates)
(uniqueness is the only requirement from the applications point of view, bad if files are overwritten).

Since the application currently doesn't care about the contents of the name, why are we?

That being said. The descriptive name is really important in troubleshooting. Sadly we do some troubleshooting.

jesperolsson-se:
This change was asked for in #842. If you wish to change or remove the `Name` functionality, please create such a ticket. Discussing this in PR comments is off-topic and benefits no one.

Unless there are any problems here and now with my code changes, please approve and merge this PR.

npetzall:
Wasn't clear that it was related to that issue. 

The concatenation was suggested early in the issue, removal was also suggested but later. If DefaultName was removed would the concatenation still be needed? Seems like Implied becomes a natural leaf.

jesperolsson-se:
Discounting `DefaultName`, we still have three leaves: `Degree`, `Implied` and `Sized`. So `Concatenation` would still be needed, yes.

npetzall:
```java
final Name name = new Concatenation(
  new Sized(compact),
  new Implied(includeImplied),
);
```
vs.
```java
final Name name = new Sized(
  compact,
  new Implied(includeImplied)
);
```

```java
final Name name = new Concatenation(
  new Degree(twoDegreesOfSeparation),
  new Implied(includeImplied)
);
```
vs.
```java
final Name name = new Degree(
  twoDegreesOfSeparation,
  new Implied(includeImplied)
);
```

jesperolsson-se:
With these suggestions, value and arrangement would be intermixed. That makes each object more complex. Unnecessarily so, unless you see some benefit other than having one class fewer. (Which itself, I would argue pretty strongly, is a drawback.)

npetzall:
So after discussions I'm getting ahead of myself. For Name composition is a better solution since there is no added functionality,
Decorator would possibly be a better pattern for building the an interactive graph. BUT, we aren't building a interactive graph yet.

Issue #849, opened 2022-04-27
Upgrade Dockerfile Java to latest?


scottjmaddox:
Is there any interest in upgrading the Dockerfile Java version to the latest, e.g. `openjdk:19-alpine3.15`? We're looking to use this in CI, but we're concerned about using such an old version of java. I've built an image from `openjdk:19-alpine3.15` with the pre-built `.jar` file, and it appears to work just fine. I'm happy to open a PR, if so.

## Expected Behavior
N/A

## Current Behavior
N/A

## Possible Solution
Change `FROM openjdk:8u111-jre-alpine` to `FROM openjdk:19-alpine3.15` in `Dockerfile`.

## Steps to Reproduce (for bugs)
N/A

## Context
Our IT dept. has concerns over the version of Java being used, and would like to use a more recent version.

## Your Environment
The docker image.

npetzall:
Jdk19 feels a bit too edgy.

We build and test with 11 but have compatibility with 8.

Going to 11 should work. Going above will probably need more work since the Javascript engine is removed.

But then again, since graphviz is installed there is no need to use vizjs.

Latest LTS is jdk17. So a pull request for 17 would be appreciated. There is a another issue with concerns related to alpine.

scottjmaddox:
PR is up: https://github.com/schemaspy/schemaspy/pull/850

PR #850, opened 2022-04-28
Dockerfile: openjdk:17-alpine3.14

scottjmaddox:
Closes https://github.com/schemaspy/schemaspy/issues/849 by upgrading the Dockerfile to Java 17 (i.e. `openjdk:17-alpine3.14`).


npetzall:
Tested with Mysq, Mssql, Pgsql so, looks good to me and as mentioned in issue vzjs isn't working as per expectation.

PR #875, opened 2022-05-20
refactor(dot): Split loop

jesperolsson-se:


npetzall (changes requested):
Both loops have a condition that the cousin should be missing from tablesWritten.
When the add is directly after it show that it's guard.

Moving the add sort of obfuscates this.

In my opinion it would be better to create a list of cousins that haven't been written.
Then loop over the missing twice and then add the missing to the written list.

Adding of the missing could be done either directly after the list has been created or at the end statement block. Since the creation of missing and adding would be in the same scope.

jesperolsson-se:
I also think that clarity would be improved by creating the filtered list. But I don't see your opinion on moving `add` - why it wouldn't be okay in the former case when it is in the latter. Could you elaborate?

npetzall:
Could you add example of which former case you're referring to?

jesperolsson-se:
> Could you add example of which former case you're referring to?

The way the PR looks right now. With expressing "missing" several times instead of storing in a temporary variable.

npetzall:
> The way the PR looks right now. With expressing "missing" several times instead of storing in a temporary variable.
Becomes a bad example when/if you force push again.

```
        for (Table relatedTable : relatedTables) {
            Set<Table> cousins = cousinsOf(relatedTable, skippedImpliedConstraints);

            for (Table cousin : cousins) {
                if (!tablesWritten.contains(cousin)) {
                    final Set<Edge> edges = new PairEdges(cousin, relatedTable, false, includeImplied).unique();
                    allCousinEdges.addAll(edges);
                }
            }

            for (Table cousin : cousins) {
                 if (!tablesWritten.contains(cousin)) {
                    final DotNode node = new DotNode(cousin, false, new DotNodeConfig(), dotConfig);
                    nodes.put(cousin, node);

                    tablesWritten.add(cousin);
                }
            }

            allCousins.addAll(cousins);
        }
```

So reasoning is as follows. 
You need to read complete code blocks to be certain about what a code block does.
The fist loop is a code block, the if is a code block. (expressions are mentally part of it's child block)

So with the code above, the first loop doesn't convey the guarding. The second does, so the concept of guarding is split over two different code blocks.

We also create the temporal aspect as was saw in the first commit that has been force-pushed away.

By creating the missing list in one code block and then adding them to tableswitten in the same code block, keeps it in the same code block or scope. If add is in the beginning or in the end is a bit about semantics and if you are reliant on the guarding, which you shouldn't be since that has been taken care of with the missing list.

Still needs to be added due to the parent loop, I assume...

npetzall:
A summy would be something like it was is one code block, want to split loop, I need to hoist the gurad so that the guard is still in the same code block.

jesperolsson-se:
Okay, so if I'm understand correctly, another split would have made it less obfuscated.

```
for (Table relatedTable : relatedTables) {
    Set<Table> cousins = cousinsOf(relatedTable, skippedImpliedConstraints);

    for (Table cousin : cousins) {
        if (!tablesWritten.contains(cousin)) {
            final Set<Edge> edges = new PairEdges(cousin, relatedTable, false, includeImplied).unique();
            allCousinEdges.addAll(edges);
    }

    for (Table cousin : cousins) {
        if (!tablesWritten.contains(cousin)) {
            final DotNode node = new DotNode(cousin, false, new DotNodeConfig(), dotConfig);
            nodes.put(cousin, node);
        }
    }
    
    for (Table cousin : cousins) {
        if (!tablesWritten.contains(cousin)) {
            tablesWritten.add(cousin);
        }
    }

    allCousins.addAll(cousins);
}
```

I guess what I'm getting at is that it seems more accurate to me to add the cousin to `tablesWritten` after we've operated on it. Anyways, I'll include your suggested refactoring in this PR.

npetzall (approved):

npetzall:
Feels very subjective, I would say one might obfuscate it even more.

The initial code before the split was self-contained. The split only added complexity due to the guarding. But as I said, hoisting the guarding sort of removes the problem.

But then again the duplication of the loops seems strange to begin with.

PR #879, opened 2022-05-26
chore: remove excluded columns from writestats

npetzall:


src/main/java/org/schemaspy/output/dot/schemaspy/DotSummaryFormatter.java
```diff
@@ -115,13 +112,4 @@ private boolean writeRelationships(Database db, Collection<Table> tables, boolea
         
         return wroteImplied;
     }
-
-    private static void markExcludedColumns(Map<Table, DotNode> nodes, Set<TableColumn> excludedColumns) {
-        for (TableColumn column : excludedColumns) {
-            DotNode node = nodes.get(column.getTable());
-            if (node != null) {
-                node.excludeColumn(column);
```
jesperolsson-se, line 123: Can this `DotNode` method be removed?

src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java
```diff
@@ -75,6 +75,7 @@ public DotNode(Table table, boolean fromRoot, DotNodeConfig config, DotConfig do
         this.dotConfig = dotConfig;
         this.path = createPath(fromRoot);
         this.columnSpan = config.showColumnDetails ? "COLSPAN=\"2\" " : "COLSPAN=\"3\" ";
+        this.excludedColumns = table.getColumns().stream().filter(TableColumn::isExcluded).collect(Collectors.toSet());
```
jesperolsson-se, line 78: Avoid operations other than assignment in the (primary) constructor. I'm fine if you do this in a secondary constructor, but if `excludeColumn` can be removed (see other comment) please also consider moving this to its own method.

src/main/java/org/schemaspy/output/dot/schemaspy/DotSummaryFormatter.java
npetzall: Yes it can, only usage was in test.

src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java
npetzall: I was trying to limit the scope.
But it might not be that big of a creep from removing excluded columns from WriteStats.
DotNode has this usage of excludedColumns:
```
if (excludedColumns.contains(column))
```

Which could be changed to
```
if(column.isExcluded())
```
Which would remove `excludedColumns` completely.

npetzall: But for the constructor "assignment".
I was initially stuck between a few different solutions. Having it as constructor arg, method in Table or going for a more media/medium approach and using the exclude Column but as withExcludedColumns.

But decided that keeping it as little as possible would leave better room for the usage analysis.

src/main/java/org/schemaspy/output/dot/schemaspy/DotSummaryFormatter.java
jesperolsson-se: Okay, then please do so

src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java
jesperolsson-se: Testability starts to suffer quickly when we add logic to the primary constructor, so I would like to not see that in this PR.

Regarding which approach we want to transition to, I agree that that would be another PR

src/main/java/org/schemaspy/output/dot/schemaspy/DotSummaryFormatter.java
npetzall: Since this comment and the second are related to the analysis and refactoring of DotNode I've created new issue for this #884

jesperolsson-se (approved):

src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java
npetzall: I totally agree constructors should be simple.

I've created a cleanup issue to analys and fix DotNod in regards to excludedColumns #884.

If the scope is retained in #884 hopefully more issues are created since one could argue that DotNode is strange in both representing Table and Columns.

Which in it self makes the testing harder.

PR #894, opened 2022-06-01
chore: writestats consolidate counting

npetzall:


src/main/java/org/schemaspy/view/WriteStats.java
```diff
@@ -27,22 +27,17 @@
  * @author John Currier
  */
 public class WriteStats {
```
jesperolsson-se, line 29: We might as well rename the class now, which allows us to (re)use it more generally. It's a `Tally`, maybe?

src/main/java/org/schemaspy/view/WriteStats.java
```diff
@@ -27,22 +27,17 @@
...
-            ++numViews;
-        else
-            ++numTables;
+    public void wrote() {
+        written++;
     }
 
-    public int getNumTablesWritten() {
-        return numTables;
+    public void wrote(int number) {
```
jesperolsson-se, line 36: This method is introduced in this PR, but nothing is using it?

src/main/java/org/schemaspy/view/WriteStats.java
jesperolsson-se, line 29: Add unit tests.

npetzall: Used LongAdder jdk class

npetzall:
Not out of the woods, but it's replaced with LongAdder and it's not used in Summary which is sort of nice.

PR #898, opened 2022-06-04
refactor(dot): Remove static

jesperolsson-se:


npetzall:
There is a bigger issue that has been discovered by this refactoring. #899

PR #916, opened 2022-06-12
chore: all orphans in a single diagram

npetzall:


npetzall:
Might be explain better with a project.
But this is the intended story.
1. Orphans as a single diagram
2. All diagrams located next to their html pages (links will be same for svg and png)
3. Remove any knowledge about relativeLinks (not used, knowledge of svg vs png reduced)
4. Remove any knowledge about svg vs png (return DiagramElement instad of DiagramResults)

jesperolsson-se (approved):
To me, it seems the use of "Diagram" is a bit conflated in the code base, especially given all the "factories", "results", "produces", etc. Please have an extra look at `OrphanDigram` and `DiagramElement` to see if those are the best names. However, this is optional as the root problem is outside the scope of this PR.

npetzall:
I stopped a bit short just to avoid scope creep.

Since changes to all of this will touch a lot of files.

But we can have a discussion about this.

npetzall:
Naming discussion in #918

Issue #904, opened 2022-06-07
Create diagrams for table according to configuration


npetzall:
The current solution creates up to 4 diagrams separating real and implied.
Implied still contains the same content as real, so there is some redundancy.
There is an application argument to skip implied which should be used.  

Create diagrams according to -degrees which is default 2.
Include implied according to -noimplied which is default false.

Issue #900, opened 2022-06-06
Investigate if moving diagrams can reduce the complexity for links


npetzall:
Links in DotNode är complex due to the handling of svg and png.
https://github.com/schemaspy/schemaspy/blob/90ac4f8d4552337f203a5ca843526c58bd5e6e00/src/main/java/org/schemaspy/output/dot/schemaspy/DotNode.java#L78-L87

When SVG is used all links are relative to the SVG, when PNG is used they are relative to the page that they are loaded in.

It might be possible to reduce the complexity by placing SVG/PNG in the same directory as the html page.
in `/tables/` `tableA.html`, `tableA.dot`, `tableA.(png/svg)`

Investigate if this is true and if true, make the change.

This would remove the `dotConfig.useRelativeLinks()`

npetzall:
This would move Summary diagrams to baseDir, table diagrams would move to tables dir, so far so good.

However orphans would also need to move to baseDir and that might be multiple diagrams. To avoid this #877 could be implemented. then there would be one additional and not X number of orphans in the baseDir. Which would be nice. We should or probably should have the 1st degree for orphans so that we always have a visual representation of table in the table page.

npetzall:
With #877 this is possible will go ahead with implementation

PR #917, opened 2022-06-12
Refactoring

jesperolsson-se:


src/main/java/org/schemaspy/output/diagram/OrphanDiagram.java
```diff
@@ -0,0 +1,23 @@
+package org.schemaspy.output.diagram;
+
+import java.io.File;
+
+public class OrphanDiagram {
```
npetzall, line 5: This is not used so can be removed

src/main/java/org/schemaspy/output/diagram/SummaryDiagram.java
```diff
@@ -0,0 +1,27 @@
+package org.schemaspy.output.diagram;
+
+import java.io.File;
+
+public class SummaryDiagram {
```
npetzall, line 5: Seems to be missing some abstraction here since this and TableDiagram are identical

npetzall (changes requested):
There are some minor issues, but this work is going to overlap with #900 and #918 

It might also overlap with #904.

src/main/java/org/schemaspy/output/diagram/SummaryDiagram.java
npetzall: Might be good to look at #918 and also DiagramElement, would be nice if we could remove DiagramResult.

jesperolsson-se: Correct (discounting the `DiagramException` message). Together with `OrphanDigram`, they existed in parallel in `DiagramFactory`. The intention of this refactoring was to eliminate that class - so I propose that introducing an abstraction would be its own PR.

src/main/java/org/schemaspy/output/diagram/OrphanDiagram.java
jesperolsson-se: On closer inspection, this class is actually used. `MustacheOrphanDiagramFactory` wasn't changed in above-mentioned PR.

npetzall: Sorry `MustacheOrphanDiagramFactory` isn't used either.

jesperolsson-se: Removing it as part of this PR would make the git history more difficult to understand. I suggest we keep the current scope, merge and create a new PR to remove the dead code.

Issue #918, opened 2022-06-12
Naming discussion - Diagram


npetzall:
We recently introduced Graph as the dot-format abstraction
Then we have Diagram which transform/generate a visualization of the dot-format (example svg or png).

With that understanding, we could rename DiagramProducer to Renderer.

We continue in the line of [DiagramElement](https://github.com/schemaspy/schemaspy/blob/b5a47ad5169a23a53478187b3c05045fced1127c/src/main/java/org/schemaspy/output/html/mustache/DiagramElement.java)

So we have something like this, DiagramElement has been renamed to Diagram, should be an interface.
So wondering if Diagram is the interface and GraphDiagram is the implementation.

```
Diagram diagram = new Diagram(Graph graph, Renderer renderer, Path outpurDir, String name);
```

Example:
```
Diagram summaryFirstDegree = new Diagram(new FirstDegree(). renderer, outputDir, "relationships.1degree");
Diagram summarySecondDegree = new Diagram(new SecondDegree(). renderer, outputDir, "relationships.2degree");
```
Then during creation of HtmlPage:
```
.addToScope("diagram", summaryFirstDegree.html())
.addToScope("diagram", summarySecondDegree.html())
```

Diagram could be responsible for setting extensions dot for dot-file and then what ever it wants for the output. (should probably correspond to format)

Note: HtmlPage, PageData.Builder are out of scope. Then can be changed later. This is only for Diagram.

npetzall:
If we want to talk on a feature level Diagram is the feature not Graph, but Graph is part of the feature, so is the renderer.
So they should live under the same package.

Diagram is also a feature of html report, so it should be moved inside of output.html

I think this would make it easier to reason about naming/feature and placement of things.

PR #920, opened 2022-06-13
Renaming diagram

npetzall:


npetzall:
#918 
Feature Diagram, components of Diagram: Graph, Renderer.
Feature Html output, components of html output: Diagram. Html

src/main/java/org/schemaspy/output/diagram/graphviz/GraphvizDot.java
```diff
@@ -150,7 +150,7 @@ private String initRenderer() {
         return EMPTY_RENDERER;
     }
 
-    public String getImplementationDetails() {
+    public String implementation() {
         return "Graphviz dot " + getGraphvizVersion().toString();
     }
```
jesperolsson-se, line 155: This is not what I expected when I read the method name and JavaDoc. Try to find a better name, perhaps along the lines of "description"

jesperolsson-se (approved):

src/main/java/org/schemaspy/output/diagram/graphviz/GraphvizDot.java
npetzall: I sort of thought.

What's the implementation?  Well implementation(), Graphvis dot 3.0.0 or Vizjs 1.18

But implementation might be a loaded word, and there might an expectation that implementation() should return an implementation which is sort of strange.

Renderer renderer = RendererStrategy.implementation()
Renderer renderer = RendererStrategy.renderer();

We can't call it "name" since it's name and version.
We might call it "identity".

The strange thing with description is that I feel that it should be more descriptive.

With the method name change the javaDoc should be updated.

name options:
* description - Textual description of implementation
* details  - Textual details about the implementation
* implementation - Textual representation of implementation
* information - Textual information about the implementation

about or description both feel like the should be more, not only name and version.

The usage is informational, in the lines of "Rendered by ....."
So "name" might not be that far of, neither are "description" or in my opinion "implementation"

or just more or less revert to implementationDetails or "details", since it can only be invoked on implementations.

npetzall: https://github.com/schemaspy/schemaspy/blob/c5a1100fd5a7139cee183a820a19f82e07e52311/src/main/java/org/schemaspy/SchemaAnalyzer.java#L321-L327

That last line should be changed during the rename.

But that's the only usage.
`writeInfo("renderer", renderer.description(), htmlInfoFile);`
`writeInfo("renderer", renderer.name(), htmlInfoFile);`
`writeInfo("renderer", renderer.implementation(), htmlInfoFile);`
`writeInfo("renderer", renderer.details(), htmlInfoFile);`
`writeInfo("renderer", renderer.implementationDetails(), htmlInfoFile);`
`writeInfo("renderer", renderer.identity(), htmlInfoFile);`

As you might have noticed I'm a bit lost.

jesperolsson-se: What makes you feel lost? All the alternatives you came up with LGTM. There's no need to look for "perfect" solutions and I bet we can always find some ambiguity if we over-analyze. Could the problem lie in the name, `Renderer` i.e., it sounds like a functional interface (compare `Engine`)?

> But implementation might be a loaded word, and there might an expectation that implementation() should return an implementation which is sort of strange.
>
> Renderer renderer = RendererStrategy.implementation()
> Renderer renderer = RendererStrategy.renderer();

Correct, this was what tripped me up.

> We can't call it "name" since it's name and version.

[Full name](https://en.wikipedia.org/wiki/Personal_name), maybe?

> The strange thing with description is that I feel that it should be more descriptive.

What information would that be (that somehow isn't details)?

PR #921, opened 2022-06-13
Refactor mustachesummary

npetzall:


src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -328,7 +328,12 @@ private void generateHtmlDoc(Config config, boolean useVizJS, ProgressListener p
         progressListener.graphingSummaryProgressed();
 
         boolean showDetailedTables = tables.size() <= config.getMaxDetailedTables();
-        final boolean includeImpliedConstraints = commandLineArguments.withImpliedRelationships();
+
+        // getting implied constraints has a side-effect of associating the parent/child tables, so don't do it
+        // here unless they want that behavior
+        List<ImpliedForeignKeyConstraint> impliedConstraints = new ImpliedConstraintsFinder(tables, commandLineArguments.withImpliedRelationships()).find();
```
jesperolsson-se, line 335: `ImpliedConstraintsFinder` should not take a boolean. Instead, the command line argument should determine what your object graph looks like.

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -331,6 +331,7 @@ private void generateHtmlDoc(Config config, boolean useVizJS, ProgressListener p
 
         // getting implied constraints has a side-effect of associating the parent/child tables, so don't do it
         // here unless they want that behavior
+        boolean hasRealConstraints = !db.getRemoteTables().isEmpty() || tables.stream().anyMatch(table -> !table.isOrphan(false));
```
jesperolsson-se, line 334: Second commit introduces distance between the comment and the code it concerns. Move `hasRealConstraints` or update the comment

jesperolsson-se (changes requested):

src/main/java/org/schemaspy/mutation/ImpliedConstraint.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Search supplied Collection of tables and add ImpliedForeignKeyConstraints
+ * when a childColumn doesn't have a parent and finds a primary key that
+ * matches based on name and type.
+ */
+
+public class ImpliedConstraint implements Mutation<ImpliedForeignKeyConstraint> {
+
```
jesperolsson-se, line 19: No constructor means that nothing is encapsulated i.e., [the objects are not unique](https://www.yegor256.com/2014/11/20/seven-virtues-of-good-object.html#3-he-is-unique).

src/main/java/org/schemaspy/mutation/ImpliedConstraint.java
```diff
@@ -0,0 +1,121 @@
...
+    // getting implied constraints has a side-effect of associating the parent/child tables, so don't do it
+    // here unless they want that behavior
```
jesperolsson-se, line 21: Misplaced comment

src/main/java/org/schemaspy/mutation/Mutation.java
```diff
@@ -0,0 +1,19 @@
...
+import org.schemaspy.model.Table;
+
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Abstraction for mutators that will modify the model.
+ * Model since a table has access to the whole object graph.
+ */
+public interface Mutation<T> {
```
jesperolsson-se, line 12: Redesign. Even though its name indicates otherwise, this interface is still a description of what the object does instead of what it is. (If the intent is to model a [mutation](https://en.wikipedia.org/wiki/Mutation), `mutate()` doesn't make sense.)

What are you trying to achieve that wouldn't be covered by a different implementation of an existing domain object? For example, a database with some extra information/different behaviour is probably still a database.

src/main/java/org/schemaspy/mutation/Mutation.java
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Perform mutation on tables
+     * @param tables that should be mutated
+     * @return list of mutations
+     */
+    List<T> mutate(Collection<Table> tables);
```
jesperolsson-se, line 18: Remove parameter. A better design is to let implementors encapsulate the tables if they are needed.

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -328,7 +332,14 @@ private void generateHtmlDoc(Config config, boolean useVizJS, ProgressListener p
         progressListener.graphingSummaryProgressed();
 
         boolean showDetailedTables = tables.size() <= config.getMaxDetailedTables();
-        final boolean includeImpliedConstraints = commandLineArguments.withImpliedRelationships();
+
+        // getting implied constraints has a side-effect of associating the parent/child tables, so don't do it
+        // here unless they want that behavior
+        List<ImpliedForeignKeyConstraint> impliedForeignKeyConstraints = impliedConstraint.mutate(tables);
+        boolean hasImpliedConstrains = !impliedForeignKeyConstraints.isEmpty();
```
jesperolsson-se, line 339: Please clarify if able, preferably with self-documenting code. My first interpretation was that the comment belongs with the command-line parsing above and that we perform the calculations prematurely, but perhaps we are dependent on the side effects?

jesperolsson-se (changes requested):

src/main/java/org/schemaspy/mutation/ImpliedConstraint.java
npetzall: It's more or less extraction of a static method.
So there was no encapsulation and no encapsulation has been added.

src/main/java/org/schemaspy/mutation/ImpliedConstraint.java
npetzall: Misplaced or redundant, should be removed.

src/main/java/org/schemaspy/mutation/Mutation.java
npetzall: Mutator, but I'm actually wondering if this is something we want.

Database isn't immutable, would feel better if it was.
All the Mutators/Augmentors/Enricher if we can call them that. Needs to much knowledge of the internals, so the Database interface/object becomes to exposed.

Well the database is a mess as it is, and simply cloning will just retain the mess.

But something like this.
```
Databsae database = 
  new DbWithImpliedConstraints(
    new DbWithRailsConstraints(
      new DbWithXmlAugmentation(
        new DatabaseFrom(
          new JDBCSource(connectionProperties)
         )
      )
   )
)
```

And maybe something similar to
```
new SchemaAnalyzer(
  new DbWithImpliedConstraints(
    new DbWithRailsConstraints(
      new DbWithXmlAugmentation(
        new DatabaseFrom(
          new JDBCSource(connectionProperties)
         )
      )
   ),
   new XmlMedium(),
   new JsonMedium(),
   new HtmlMedium()
).analyze();
```

So might be better to roll all of this back, and focus only on moving it out of MustacheSummaryDiagramFactory.

And figure out how to decorate the Database instead.

src/main/java/org/schemaspy/mutation/Mutation.java
npetzall: https://github.com/schemaspy/schemaspy/pull/921#discussion_r896912261
> Separate creation and execution

https://github.com/schemaspy/schemaspy/pull/921#discussion_r896915958
> `ImpliedConstraintsFinder` should not take a boolean. Instead, the command line argument should determine what your object graph looks like.

Since there is no object graph over how the application should work and I assume we want to create the graph in Main in reality, so there is no possibility of having Table as a constructor argument.

The closes to this is if in Entry point. Produce something in the line of 
```
new SchemaAnalyzer(
  new DbWithImpliedConstraints(
    new DbWithRailsConstraints(
      new DbWithXmlAugmentation(
        new DatabaseFrom(
          new JDBCSource(connectionProperties)
         )
      )
   ),
   new XmlMedium(/*.....*/),
   new JsonMedium(/*.....*/),
   new HtmlMedium(/*.....*/)
).analyze();
```

Which probably would make more sense, to have constructor with something that can produce a Database or something that can take a Medium,

jesperolsson-se: > I assume we want to create the graph in Main in reality, so there is no possibility of having Table as a constructor argument.

The design you propose (which I agree with) does this so I'm not sure what point you're trying to make. `DbWithImpliedConstraints` would have access to the decoratee who can be asked for `Tables` (or whatever contract we end up with when we get there).

npetzall: I'm only disagreeing with tables as an encapsulatee and it relates to the other comments. That a good encapsulatee doesn't exist at the moment and as such this isn't the implementation we want.

PR #923, opened 2022-06-14
Refactoring

jesperolsson-se:


npetzall (approved):

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -94,18 +94,22 @@ public class SchemaAnalyzer {
 
     private final CommandLineArguments commandLineArguments;
 
-    private final List<OutputProducer> outputProducers = new ArrayList<>();
+    private final OutputProducer outputProducer;
 
     public SchemaAnalyzer(SqlService sqlService, CommandLineArguments commandLineArguments) {
```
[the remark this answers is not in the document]
jesperolsson-se: If you want to do that, I don't mind. The object isn't sufficiently well-defined for me to be able to argue about trade-offs

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -94,18 +94,22 @@ public class SchemaAnalyzer {
...
-        this.databaseServiceFactory = new DatabaseServiceFactory(sqlService);
-        this.commandLineArguments = Objects.requireNonNull(commandLineArguments);
-        addOutputProducer(new XmlProducerUsingDOM());
+        this(sqlService, new DatabaseServiceFactory(sqlService), commandLineArguments, new XmlProducerUsingDOM());
     }
 
-    public SchemaAnalyzer addOutputProducer(OutputProducer outputProducer) {
-        outputProducers.add(outputProducer);
-        return this;
+    public SchemaAnalyzer(
```
npetzall, line 103: JavaDoc

The previous behavior was that the xml output was always included but others could be added.
This new behavior is good that Xml can be replaced.

And we don't get any javadoc unless we add it.

src/main/java/org/schemaspy/SchemaAnalyzer.java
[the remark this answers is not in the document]
npetzall: Trade off in this case is missing javadoc I guess. With two different constructor and no immediate way of knowing the difference. It must be investigated, might be obvious that it's the outputproducer that is changed in some way but what way.

src/main/java/org/schemaspy/SchemaAnalyzer.java
jesperolsson-se: The object isn't sufficiently well-defined for me to be able to write helpful JavaDoc (yet). Given the state of this class, I'd say no JavaDoc is better than a bunch of
> \* @param outputProducer The outputProducer.

I'm happy to add it if I receive enough information

src/main/java/org/schemaspy/SchemaAnalyzer.java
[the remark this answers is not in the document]
jesperolsson-se: I disagree, no JavaDoc has been removed and the code is now clearer and more self-documenting. Either constructor upholds the contract that a fully serviceable object will exist after called. More parameters (or more sophisticated types) enable the caller to give a more precise specification, but that information is not essential for the object to fulfill its contract.

src/main/java/org/schemaspy/SchemaAnalyzer.java
npetzall: From my point of view

```
/**
* Constructed with new XmlProducerUsingDOM @see OutputProducer @see new XmlProducerUsingDOM
*/
```
For the Primary constructor
```
/**
* Constructed with supplied OutputProducer @see OutputProducer
* /
```

Would been sufficient as a start.

jesperolsson-se: I disagree with the first one and the second one is just a reformulation of what I objected to in my previous comment.

If I use a constructor that doesn't have `OutputProducer`, then its JavaDoc should not mention it and definitely not any of its concrete implementations. It's neither needed nor relevant and it reduces maintainability and confuses the user of the constructor.

But perhaps your concern isn't with the JavaDoc, but with the design? Neither constructor is likely to exist when `SchemaAnalyzer` has been fully refactored - and better designs will be discovered as we continue.

npetzall: > If I use a constructor that doesn't have OutputProducer, then its JavaDoc should not mention it and definitely not any of its concrete implementations. It's neither needed nor relevant and it reduces maintainability and confuses the user of the constructor.

Well, it should be that way but it isn't. That's why initially asked about the removal.

The removal of the secondary would be  better. Since the "default" isn't some small thing.

Issue #247, opened 2018-02-09
Remove usage of ProgressListener.recoverableExceptionEncountered


npetzall:
There's a single implementation and it does light formatting.
Also recoverableException has nothing to do with Progress.

Causes issues when progresslistener is mocked in tests.

npetzall:
They will be removed if the new progress system with a progressbar is merged.

npetzall:
Closed by fbaddce

PR #924, opened 2022-06-17
Refactoring

jesperolsson-se:


npetzall (changes requested):
ProgressListener is bad, most of the methods that "render"/write to console har only invoked if Html.
So ConsoleProgressListener doesn't do anything with the console except for 1 method that has something to do with exceptions.

Two things that could be done, 
1. skip the whole if htmlEnabled. It will almoste never be called if it's disabled. Has more with verbose than html, but only used when doing html. Strange stuff.

But the responsibilities could be divided. 
ConsoleProgressListener.java -> Tracked.java
Html -> Console.java

src/main/java/org/schemaspy/model/ConsoleProgressListener.java
```diff
@@ -33,147 +31,76 @@
...
+    }
+
+    @Override
+    public long finished(Collection<Table> tables, Config config) {
+        finishedAt = System.currentTimeMillis();
+        return finishedAt - startedAt;
+    }
+
+    @Override
+    public String recoverableExceptionEncountered(String msg, Exception exc, String sql) {
```
npetzall, line 95: Remove, this isn't progress, it not used, but declared by the interface.

src/main/java/org/schemaspy/model/ConsoleProgressListener.java
```diff
@@ -33,147 +31,76 @@
...
+    private long startedGraphingDetailsAt;
+    private long finishedAt;
 
-	@Override
-	public void gatheringDetailsProgressed(Table table) {
-		if (render) {
-			System.out.print('.');
-		}
-	}
+    public ConsoleProgressListener() {
```
npetzall, line 42: Doesn't do much with "Console", should Tracking, Timing, or Simple, Base

src/main/java/org/schemaspy/model/Html.java
```diff
@@ -0,0 +1,108 @@
...
+import org.schemaspy.Config;
+import org.schemaspy.cli.CommandLineArguments;
+
+import java.io.File;
+import java.util.Collection;
+
+/**
+ * Decorator for rendering console-based details for HTML.
+ */
+public class Html implements ProgressListener {
```
npetzall, line 12: Most methods are used by Html generating Code. But what it does is write to System.out which in it self isn't great.

System.out is StdOut so may be StdOut or Console.

src/main/java/org/schemaspy/model/Html.java
```diff
@@ -0,0 +1,108 @@
...
+        long result = origin.finished(tables, config);
+        System.err.flush();
+        System.out.flush();
+        System.out.println("Wrote relationship details of " + tables.size() + " tables/views to directory '" + commandLineArguments.getOutputDirectory() + "' in " + result / 1000 + " seconds.");
+        System.out.println("View the results by opening " + new File(commandLineArguments.getOutputDirectory(), "index.html"));
+        return result;
+    }
+
+    @Override
+    public String recoverableExceptionEncountered(String msg, Exception exc, String sql) {
```
npetzall, line 105: Should be removed from interface and thus from here.

jesperolsson-se (commented):
The review comments are about pre-existing implementation. I'd be happy to resolve them in new PRs (after this one is merged). Do you have any concerns regarding the refactoring?

npetzall:
The naming issue is introduced.

npetzall (approved):

npetzall:
The removal is mentioned in #247

PR #926, opened 2022-06-22
refactor(analyzer): Extract object

jesperolsson-se:


src/main/java/org/schemaspy/LayoutFiles.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Enumeration;
+import java.util.Objects;
+
+/**
+ * Represents the contents of the layout folder.
+ */
+public class LayoutFiles {
+
+    private final ClassLoader classLoader;
+    private final String DOT_HTML;
```
npetzall, line 20: Either take a FileFilter or rename to something better.

`DOT_HTML` was a constant now it isn't, it's used in a suffix Filter, and it was an extension.

src/main/java/org/schemaspy/LayoutFiles.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.io.File;
+import java.io.FileFilter;
+import java.io.IOException;
+import java.net.URL;
+import java.util.Enumeration;
+import java.util.Objects;
+
+/**
+ * Represents the contents of the layout folder.
```
npetzall, line 15: If it represents the content of the layout folder, why should it know about the output dir?

Or are we suggesting outputDir is the layout folder?

npetzall (changes requested):

npetzall:
Feels like this could be more generic. We are talking more or less of a filset that we want to copy.

So an interface FileSet a method copy(Path destination).

Implementation Layout(don't even know if layout is correct, it's assests for HTML ouput/report).

Could probably be devided into even smaller objects.

new FileSet(new Collector(new FileFilter))

jesperolsson-se (commented):
Let's keep the scope small. Making the objects and smaller are both big enough to warrant PRs of their own.

src/main/java/org/schemaspy/LayoutFiles.java
npetzall: Let me rephrase, at some point it could know about the outputDir. But having it part of the constructor makes this, not a representation of Layout Content, but a CopyFilesTask.

We create this object twice, they differ in the outputdir, not in the layout content. But that's probably a discussions for the later PR's.

npetzall (approved):

PR #927, opened 2022-06-22
Work towards SchemaSpy without Spring (Boot)

onkobu:
- fixture replaces ApplicationContext
- Integration tests do not need any context anymore
- bump versions/ make versions explicit in pom.xml
- drop configuration classes


pom.xml
```diff
@@ -198,10 +198,20 @@
...
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-core</artifactId>
+            <!-- Without Spring Boot this must be explicit -->
+            <version>2.2</version>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
+            <!-- Without Spring Boot this must be explicit -->
+            <version>4.6.1</version>
```
npetzall, line 208: Version for mockito is defined in depencyManagement.

onkobu: Moved it. Nevertheless I don't see any advantage of the split between dependencyManagement and dependencies without child modules. The few and isolated pinned versions for test dependencies seem straightforward. Thus they could easily end up in a single <dependencies> block.

npetzall:
We done some work on multiple fronts and this PR was used as inspiration for #1246.

We have moved to java 17, and there is still things to do regarding the packaging. Since we still rely on spring-boot.

So I think we are on par with this PR.

Thank you for you contribution, it acted as motivation and inspiration for the changes we have made.

PR #931, opened 2022-06-29
Refactoring

jesperolsson-se:


src/main/java/org/schemaspy/LayoutFiles.java
```diff
@@ -15,21 +11,17 @@
...
-    public LayoutFiles(ClassLoader classLoader, FileFilter filter, File outputDir) {
+    public LayoutFiles(ClassLoader classLoader) {
         this.classLoader = classLoader;
-        this.filter = filter;
-        this.outputDir = outputDir;
     }
 
     /**
-     * Copy the files - sans template .html files - to object's destination.
+     * The layout files - sans template .html files.
```
npetzall, line 20: The sans part isn't true anymore, the filtering is done by the ResourceWriter

src/main/java/org/schemaspy/LayoutFiles.java
```diff
@@ -15,21 +11,17 @@
...
      *
-     * @throws IOException when not possible to copy layout files to outputDir
+     * @throws IOException when not possible to retrieve layout files
      */
-    public void prepare() throws IOException {
+    public URL files() throws IOException {
```
npetzall, line 24: It's strange with `files()`and only returning a single URL.

I'm guessing we probably should be talking about `LayoutFolder`

npetzall (changes requested):

src/main/java/org/schemaspy/LayoutFiles.java
jesperolsson-se: If correct, then we can remove most or all of `ResourceWriter`

npetzall: My issue was name implies plural, return is singular.

Would still be a URL just that the method would be folder and the object be a LayoutFolder. Not sure how that would impact ResourceWriter.

As mentioned in another PR, one responsibility would be to provide a collection of files, one filter that collection and a third writing them. ResourceWriter has lots of responsibilities now.

src/main/java/org/schemaspy/SchemaAnalyzer.java
```diff
@@ -114,19 +115,38 @@ public Database analyze(Config config) throws SQLException, IOException {
         if (commandLineArguments.isHtmlEnabled()) {
             progressListener = new Console(commandLineArguments, progressListener);
         }
-        
+
+        LayoutFiles layoutFiles = new LayoutFiles(SchemaAnalyzer.class.getClassLoader());
```
npetzall, line 119: Shouldn't we make it a field to avoid passing it around?

src/main/java/org/schemaspy/LayoutFiles.java
jesperolsson-se: If you check the usages of `ResourceWriter` you will find none other than with `LayoutFiles` even though it has code for handling a file, a folder and jar i.e., we have dead code.

npetzall (changes requested):

npetzall (approved):

PR #936, opened 2022-07-07
refactor(analyzer): Extract object

jesperolsson-se:


npetzall (changes requested):
The names are all really strange.
They seem to be derived from `Markdown#registryPage()`
https://github.com/schemaspy/schemaspy/blob/31a69291ae8ab4bf49a790e5236491351c2e0718/src/main/java/org/schemaspy/util/Markdown.java#L65-L67

Extraction of this method is only done since the "registryPage" method in Markdown is "incompatible" with the data we have at hand.

So seems that we are extracting an Object due to an error in Markdown.

Registration of a tableName and tablePath truly only acts as a condition for when a "reference" in markdown is rendered as a table link.
https://github.com/schemaspy/schemaspy/blob/31a69291ae8ab4bf49a790e5236491351c2e0718/src/main/java/org/schemaspy/util/Markdown.java#L69-L71
Well it's even a bad one...., if it renders a valid link.
https://github.com/schemaspy/schemaspy/blob/31a69291ae8ab4bf49a790e5236491351c2e0718/src/main/java/org/schemaspy/util/Markdown.java#L100-L104

Markdown is static which means if Schema A contains a real  table A this would be added to the registry. If then Schema B contains a .... logical table A it would not be added, but it would already be "registered" by Schema A. It's a bug.

But making Markdown non static, having the FileNameGenerator and a list of "approved" table names for the current "report(schema)", would remove the object extraction, the badly named Markdown.registryPage and fix that bug. 

We would still be required to fix the bug regarding having "null" in path if the Markdown.pagePath returns null.

npetzall:
To clarify, I don't think we should extract the method to an object, I think we should fix Markdown instead.
If you want to improve your commit statistics I can add this, so that you can remove it later.

However I'm gonna add two Issues, the static bug thing and the null path.

npetzall:
If the goal is to remove clutter, just refactor the method in Markdown.

And leave the refactoring to make it non-static to later.

npetzall (approved):

PR #940, opened 2022-07-10
Refactoring

jesperolsson-se:


npetzall:
What about having decorator instead of a static registry. 

new Markdown(new ReferenceLinks(text, path, tables)).html()

jesperolsson-se (commented):
I want to keep the scope small for each PR, so I won't do that in this one. But if you think that would be a good idea, send up a new one and I can review.

npetzall:
If PageRegistry wasn't extracted the pagePath could have been private?

I think the PageRegistry and ReferenceLink are part of the same object which makes the separation strange.

jesperolsson-se (commented):
A private method in itself is a pretty good indicator that the object is not properly decomposed. On the contrary, I think the ideal would be to break this down even further.

If you look at the callers, you'll also find that these are two different responsibilities. `SchemaAnalyser` registers the pages (statically) but never asks for markdown/html. Instead, that is done in seven different "view" classes.

Still, I think you're correct in that there's an better refactoring where we don't have to use a registry but from what I can tell right now, that one's inaccessible behind the current architecture.

npetzall:
> A private method in itself is a pretty good indicator that the object is not properly decomposed. On the contrary, I think the ideal would be to break this down even further.

I more or less agree with you, but rather that I only wanted to point out that the registry might not be needed. It's populated "once" which suggests that It's actually a constructor. It's used inside something and only that something.

> If you look at the callers, you'll also find that these are two different responsibilities. SchemaAnalyser registers the pages (statically) but never asks for markdown/html. Instead, that is done in seven different "view" classes.

Since SchemaAnalyser has far too many responsibilities inside, we might need to logically break it down to `generateHtmlDoc` and within `generateHtmlDoc`, as you pointed out, only views use it (indirectly, it uses Markdown which uses it).

Views are constructed in the `generateHtmlDoc` so the static methods only avoid a constructor arg of object created in the same method body. Which makes that static feel really lazy.

So to keep the scope small, I wouldn't extract the registry, I would rather wait and have a look at the whole concept.
It has a method that is a constructor, it's used by one single thing. It should be that objects constructor.

Markdown has too many responsibilities at least reference links and calling the htmlrenderer. So they should be separated.

The main argument is that it has a method that is a constructor.

jesperolsson-se (commented):
Okay, yes! That makes perfect sense. Thanks for the clarification!

PR #942, opened 2022-07-15
Refactoring

jesperolsson-se:


jesperolsson-se (commented):
> naming shouldn't be under util. It should be directly under schemaspy.

I disagree. Naming isn't endemic to SchemaSpy. In fact, I'm considering if they should be extracted as a library. More specific name objects remain at `output/dot/schemaspy/name`.

npetzall:
I think we are in essence after the same thing. It's a matter of perspective regarding org.schemaspy if it's the organization or the application.

If it was a library released by Schemaspy it would probably be org.schemaspy.naming.

But there might also exist a library that it can be replaced with. But from a microtyping perspective we still might need to add a shim, for instance if we were to use Cactoos. A shime would be to extend/wrap Text with a Name interface. (not sure about the wrap/extend, just a spontaneous idea)

jesperolsson-se (commented):
Is your comment speculation? What's keeping this PR from being merged? Since we're in agreement about the reusability aspect, I'd say `util` is an appropriate location, with potential extraction being a separate task.

npetzall:
Cactoos replacement is speculation.

Mainly its blocked like I said in the first comment that I haven't had the time to look it over.

But after the first and second comment regarding util. I realize that we have some general placement issues.

For example there are stuff in utils that shouldn't be there. We have a big chunk of Html documentation code in SchemaAnalyzer. Html documentation code is both under output, view and utils.

In general I don't want a utils package. Because things aren't both that general and that unimportant at the same time.

SchemaSpy is linear/sequential in my mind. 
Input->Stuff->Output
There is currently nothing in utils that spans the whole flow. Things should be moved from utils not into utils.

PR #954, opened 2022-08-29
Refactoring

jesperolsson-se:


src/main/java/org/schemaspy/connection/SemicolonSeparated.java
```diff
@@ -0,0 +1,39 @@
...
+    private static final String ESCAPED_EQUALS = "\\=";
+
+    /**
+     * @param properties string with key\\=value pairs separated by ; of connection properties
+     */
+    public SemicolonSeparated(String properties) {
+        this.properties = properties;
+    }
+
+    public Properties setConnectionProperties() {
```
npetzall, line 25: Why is this set? Shouldn't even be get, since it returns a new Properties.

src/main/java/org/schemaspy/connection/Filed.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param propertiesFilename file to use for connection properties
+     */
+    public Filed(String propertiesFilename) {
+        this.propertiesFilename = propertiesFilename;
+    }
+
+    /**
+     * @throws IOException if we have problems reading the file
+     */
+    public Properties setConnectionPropertiesFile() throws IOException {
```
npetzall, line 28: set seems awkward when constructing a new object and not changing any state.

src/main/java/org/schemaspy/connection/Filed.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * Properties from this file (in key=value pair format) are passed to the
+ * database connection.<br>
+ * user (from -u) and password (from -p) will be passed in the
+ * connection properties if specified.
+ */
+public class Filed {
```
npetzall, line 14: Filed is a term for archiving, this is more of a PropertiesFromFile.

Which more or less probably can be abstracted to some kind of Input.
Since the Inputstream from a file could look the same as a string made into an inputstream.

src/main/java/org/schemaspy/connection/SemicolonSeparated.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.StringTokenizer;
+
+/**
+ * Specifies connection properties to use in the format:
+ * <code>key1\=value1;key2\=value2</code><br>
+ * user (from -u) and password (from -p) will be passed in the
+ * connection properties if specified.<p>
+ * This is an alternative form of passing connection properties than by file
+ */
+public class SemicolonSeparated {
```
npetzall, line 13: I have somewhat of a feeling that if `;` was replaced with `\n` and the `\\=` was just a regular `=` we could skip our processing and just send it straight to one of the load functions of properties.

npetzall (changes requested):
If these are just half-way changes and that the goal is somewhat close to my comments let me know and I'll merge.

src/main/java/org/schemaspy/connection/Filed.java
jesperolsson-se: `PropertiesFromFile` is a bit long, but otherwise yes. What about `FromFile`?
Generalization is a later step that might not be worthwhile.

src/main/java/org/schemaspy/connection/SemicolonSeparated.java
jesperolsson-se: Happy to try that when this is merged. Or do you want to?

src/main/java/org/schemaspy/connection/SemicolonSeparated.java
jesperolsson-se: I agree but won't do everything at once. Below is whatit looked like previously.
https://github.com/schemaspy/schemaspy/blob/0f31217787f17b2cf710e1962f0df9a249eb2d77/src/main/java/org/schemaspy/ConnectionProperties.java#L33-L36

jesperolsson-se:
I guess you could call it half-way. I don't really think about refactorings in that way

src/main/java/org/schemaspy/connection/Filed.java
npetzall: Well the previous methods were setting a state, now they are not. But they retained the "set" prefix.

Which is confusing, like it's mid refactoring. I changed the method body but not the name.

src/main/java/org/schemaspy/connection/Filed.java
npetzall: It's a bit long but, the Object has multiple responsibilities so I think it's fitting.

Wanna make it shorter, remove some of it's responsibilities.

src/main/java/org/schemaspy/connection/SemicolonSeparated.java
npetzall: It can be done at a later stage, it will be an experiment to see if it possible.

npetzall (approved):

PR #976, opened 2022-10-31
refactor(con): Introduce abstraction

jesperolsson-se:


npetzall:
I might need some time to digest this. But I initially have two questions.

1, Why doesn't ConnectionPropertiesFactory implement Connection?
2, Or why doesn't ConnectionPropertiesFactory return a Connection?

I'll need to work thru my bias, regarding Connection interface, of knowing how it's currently used and that the JavaDoc is misleading at the moment, but also the existing Connection interface in JDBC. (Some kind of large scale shadowing)

But there's also the fact that how other objects would collaborate with Connection.
If I were to collaborate with Connection it would probably be to interact with the database. (This is a lot of JDBC bias)

If we for instance would go more or less the micro typing route. 
We would have ConnectionProperties as an interface that returns properties.

And from my Cactoos experience it would just be Scalar<Properties> using the new PropertiesOf.

I'm gonna sleep on it and see if it makes sense in the morning, or am I going overboard in that I want to understand it?

jesperolsson-se:
> 1, Why doesn't ConnectionPropertiesFactory implement Connection?

No real reason. If you think it makes sense to include that in this PR I can do that.

2, Or why doesn't ConnectionPropertiesFactory return a Connection?

Would be too many changes in one step. Notice how it sometimes returns `new Properties()`.

npetzall:
> > 1, Why doesn't ConnectionPropertiesFactory implement Connection?
> 
> No real reason. If you think it makes sense to include that in this PR I can do that.

If it did implement Connection would it be renamed? 

> 2, Or why doesn't ConnectionPropertiesFactory return a Connection?
> 
> Would be too many changes in one step. Notice how it sometimes returns `new Properties()`.

If feel it's sort of big changes but incomplete so it makes it harder to understand. But I'm working on it. 
`() -> new Properties()` and you would have a Connection. Might even be possible with new Semicolon...("").properties()

It's more or less implementations of the Supplier<Properties> interface is what I'm thinking. But I'm trying to figure out how I can reason about Connection and it having properties that I want to expose and why.

jesperolsson-se:
> > > 1, Why doesn't ConnectionPropertiesFactory implement Connection?
> > 
> > No real reason. If you think it makes sense to include that in this PR I can do that.
> 
> If it did implement Connection would it be renamed?

Fixed.

> `() -> new Properties()` and you would have a Connection.

Fixed.

npetzall:
Only thing is that Connection will eventually clash with Connection.

But that could be a problem for a later day, and it could be internal to Connection.

So lets see were this takes us.

PR #992, opened 2023-02-09
refactor(con): Extract password

jesperolsson-se:


src/test/java/org/schemaspy/connection/WithPasswordTest.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public final class WithPasswordTest {
+
+    /**
+     * Given a password and an origin,
+     * When the object is asked for connection properties,
+     * Then its response should include the password.
+     */
+    @Test
+    public void parse() throws IOException {
```
npetzall, line 21: Parse feels strange as testname. Can we come up with a better one?

src/test/java/org/schemaspy/connection/WithPasswordTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.io.IOException;
+import java.util.Properties;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+/**
+ * Tests for {@link WithPassword}.
+ */
+public final class WithPasswordTest {
```
npetzall, line 13: I would like to have the null case as a test, so that it forms sort of a specification. The SSO requires null handling.

npetzall (changes requested):

src/test/java/org/schemaspy/connection/WithPasswordTest.java
jesperolsson-se: I'm open for suggestions :slightly_smiling_face: 
If it helps, I mentally think "I can [test method name]"

src/test/java/org/schemaspy/connection/WithPasswordTest.java
jesperolsson-se: I add can it for now, but we should move away from the specification. Alternative is [Null Object](https://en.wikipedia.org/wiki/Null_object_pattern)

src/test/java/org/schemaspy/connection/WithPasswordTest.java
npetzall: I can SetAnActualPassword

jesperolsson-se: `readPasswords`, maybe? But I still think "parse" is clearer

npetzall: It doesn't read, it doesn't parse and it's a with password so something gets a password. Hence a password is set not in the setter way.

jesperolsson-se: Hmm, I think we're doing two (three, actually) things. The first one should probably be decomposed in a future refactoring.

1. read/parse the password from config
2. decorate origin with the password
3. `null` handling

I think you saw it from (2.) while I was focusing on (1.)

"I can decorate"?
"I can supply password"?

npetzall: It's not reading or parsing from config, that might be something that config does. At most we retrieve it from config.

Yes decorating, in a sense, it feels more like a chained command. 

Null handling is a different test case. We should only handle a single code path in a single unit test. But that's more of a glass half full or half empty.

Do we handle null or do we handle non-null?

src/test/java/org/schemaspy/connection/WithPasswordTest.java
npetzall: Or if there might be a way to avoid it. If password hasn't been specified we shouldn't have a withpassword.

If possible, not specified nothing happens, specified something happens.

NoPassword default and WithPassword when specified and Setting interface has apply(Map<String, Object> context).

Only brainstorming, not sure how I feel about it.

npetzall (approved):

Issue #993, opened 2023-02-15
Bump Spring-boot to version 2.2.11 to fix CVE-2022-27772


vunhan:
CVE-2022-27772

## Expected Behavior

To mitigate CVE-2022-27772, we need to bump Spring-boot to version 2.2.11

Detected with AWS Inspector

PR #994, opened 2023-02-15
Bump Spring-boot to v2.2.11

vunhan:
To mitigate CVE-2022-27772, we need to bump Spring-boot to version 2.2.11

Fix issue #993

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
         <argLine>-Dfile.encoding=UTF-8</argLine>
         <sonar.junit.reportPaths>target/surefire-reports,target/failsafe-reports</sonar.junit.reportPaths>
 
-        <spring-boot.version>2.1.2.RELEASE</spring-boot.version>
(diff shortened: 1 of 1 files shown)


npetzall:
This update breaks on testCompile phase in maven.

npetzall:
Fix is on the way

npetzall:
Seems that you've created a PR in a fork of a fork, so I don't have access to update it.
Could you apply the following patch
[0001-chore-test-replace-with-new-implementation.patch](https://github.com/schemaspy/schemaspy/files/10788031/0001-chore-test-replace-with-new-implementation.patch)

`git am < 0001-chore-test-replace-with-new-implementation.patch`

npetzall:
Due to lack of update, I'll try to add the patch before this, or I'll create my own PR with this commit and the fix.

npetzall:
Replaced by #1030

Issue #989, opened 2023-02-02
schemaspy-6.1.0.jar should not be installed in a path containing spaces


nausicaaconsulting:
## Expected Behavior

` java -jar ./schemaspy-6.1.0.jar -dbhelp`
Should display the list of supported connection types

## Current Behavior

If schemaspy-6.1.0.jar is installed in a folder which contains spaces in its absolute path: one receive the following message instead of the list of supported database types : 

`$ java -jar ./schemaspy-6.1.0.jar -t mssql -dbhelp

SchemaSpy generates an HTML representation of a database schema's relationships.
SchemaSpy comes with ABSOLUTELY NO WARRANTY.
SchemaSpy is free software and can be redistributed under the conditions of LGPL version 3 or later.
http://www.gnu.org/licenses/

INFO  - Starting Main v6.1.0 on BOD-PF1CCTNY with PID 12052 (C:\Users\dummy\Path with space\schemaSpy\schemaspy-6.1.0.jar started by dummy in C:\Users\dummy\Path with space\schemaSpy)
INFO  - The following profiles are active: default
INFO  - Found configuration file: schemaspy.properties
INFO  - Started Main in 1.037 seconds (JVM running for 1.511)
INFO  - Built-in database types and their required connection parameters:
ERROR - Failed to read bundled DatabaseTypes
C:\Users\dummy\Path%20with%20space\schemaSpy\schemaspy-6.1.0.jar (Le chemin dÆaccÞs spÚcifiÚ est introuvable)
INFO  - You can use your own database types by specifying the filespec of a .properties file with -t.
INFO  - Grab one out of /C:\Users\dummy\Path%20with%20space\schemaSpy\schemaspy-6.1.0.jar and modify it to suit your needs.
INFO  - StackTraces have been omitted, use `-debug` when executing SchemaSpy to see them
`

## Steps to Reproduce (for bugs)

1. install JAR file in a path with a space in it  : -dbhelp command fails
2. remove the space in your path: -dbhelp command works
3. Regret you work with OneDrive

## Context

* Database type (e.g. MySQL):
* Database driver version (is you use a special one):
* Your DDL (sample how to reproduce the issue):
* Full command including arguments how you run SchemaSpy:  java -jar ./schemaspy-6.1.0.jar -dbhelp

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39): 
* Operating System and version (e.g. Windows 10): Windows 10
* Java version: 
java version "1.8.0_361"
[...]

PR #995, opened 2023-02-15
#989 Decode path

jesperolsson-se:


npetzall:
I'll try this on windows, I'm a bit worried that, since it will hit the if loaded from !BOOT and as such it will create a new URL, it will just encode it again.

But we will see when I get to testing.

npetzall:
Tested OK.

Issue #733, opened 2020-10-16
Feature request: Move to GraalJS to replace Nashorn, Nashorn removed in Java 15.


nhoughto:
So Nashorn has been removed from Java 15 so the viz.js feature no longer works, there should be a simple change to use GraalJS to run viz.js instead. You can pull graaljs in as a dependency and don't need to mandate users actually run GraalVM.

It would increase the size of the dependencies of users though, and you may want a separate maven target to opt-in to graaljs behaviour.. if interested i can probably PR some of it..

npetzall:
Do you know if GraalJs will work on Java 8?
If we need to do two builds Java8 and Java 15 where 15 might be a bit bigger.

Haven't even tried building on 15 will be interesting.

nhoughto:
Yep GraalJS runs on Java 8

it will be the same size on both, as you just pull it in as a regular dependency, https://mvnrepository.com/artifact/org.graalvm.js/js/20.2.0

npetzall:
With size I was thinking a java8 build without and java15 build with.

But I checked and it's serviceloaded as a scriptEngine. So might be possible to change a bit of classpath and instructions to download and put in a sibling directory and it's available.

Worst case, we could run the viz.js with it's own classloader and we add ./js-*.jar to that classloader and use it.

nhoughto:
ah right yeah i just thought you'd migrate the whole thing to graaljs in all versions, then just have a maven target for schemaspy+vizjs that bundles graal and remove the nashorn code entirely..

npetzall:
Without testing i believe that the following might work
Replace the normal start 
`java -jar schemaspy-6.1.1-SNAPSHOT.jar`
With
`java -cp schemaspy-6.1.1-SNAPSHOT.jar;js-20.2.0.jar org.springframework.boot.loader.JarLauncher`

Separator between the jars is OS-dependet `:` on linux and `;` on windows.
But I admit, I haven't used java15 so don't know if they have totally changed that so it should be some kinde of module path.

I guess I have to download and test it.

nhoughto:
ah ok yep that could work

This also seems like a option? rather than solving it all yourself, let the library make graphviz work cross platform - https://github.com/nidi3/graphviz-java

npetzall:
We/I looked at graphviz-java like 2 years a go, might need to take another look.

But the complex code is the generation of dot-lang files and not executing graphviz or viz.js and graphviz-java has a model for the language.

[...]

PR #996, opened 2023-02-17
make vizjs to work with jdk15 or higher.

digitalbot:
I added nashorn dependency for jdk15 or higher.
I couldn't find the contributor rules, so if there is a way to do it, please let me know.

### changes
- add dependency nashorn
- specify nashorn script-engine

### references
https://mvnrepository.com/artifact/org.openjdk.nashorn/nashorn-core/15.4
https://stackoverflow.com/questions/65265629/how-to-use-nashorn-in-java-15-and-later

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -64,8 +64,6 @@ SchemaSpy version 6.1.0 and higher now comes with viz.js, so you will not need t
 
 For SchemaSpy version 6.1.0 or higher, simply include ``-vizjs`` as a command line argument when executing the SchemaSpy command.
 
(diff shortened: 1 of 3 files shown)


npetzall:
This would be a great fix for the javascript after java 15. This has been discussed in #733, since then nashorn doesn't requirem JPMS, it can run with out module-path which would be good. Just need to check how much it increases the size since we try to minimize it.

In #733 there is an idea about adding additional classpath so that nashorn can be "side-loaded" which would make it optional. But it's great news that it can be used without JPMS.

npetzall:
In #733 we were also thinking about having two jars. One with, and one without script engine.

19MB isn't that bad, i think without it's like 18MB. But I also noticed during the weekend that snapshot build i much bigger than the latest release. So I have some additional work...

I'll try to give it a go tonight. I'm so excited.

npetzall:
It was a wonderful experience.

@jesperolsson-se if this looks ok to you, you can rebase and merge it.

PR #1000, opened 2023-02-19
971 doc database types

npetzall:
Add sphinx extension to read from included dbtypes and add to documentation


npetzall:
Either you build it and check it, or we setup a session to go over it.
I'm not sure about the layout/placement so I need you input.

jesperolsson-se:
Will review after #1003 is resolved

npetzall:
Are you sure that 1003 isn't an issue with this PR?

jesperolsson-se:
You're correct. My bad!

jesperolsson-se (changes requested):
I'd say, let's put the database name and the argument next to each other. Optionally, put them in a table.

Also, update the README :wink:

npetzall:
Should they still be in the same page or should it be its own page?

I wonder if they are sorted, might be a good thing to have it sorted.

jesperolsson-se:
> Should they still be in the same page or should it be its own page?

Same page for now. If someone finds the current page, they will see a pretty long page that also contains niche information. I think it's quite unlikely that they would start looking for "included" somewhere else. In the future, however, I agree that its own page would be better.

> I wonder if they are sorted, might be a good thing to have it sorted.

Good suggestion, but it might be a bit complicated. I recommend getting this merged first and then look into sorting. For example, I think it's most natural to sort on the database name (not file name nor the CLI parameter), you need to consider "JTDS", "11 or later" and similar - maybe even rename some of them?

npetzall:
I've added a dbms property to databaseType and reduced the description. I also changed the css for the table after our talk.

It now looks like
image

I'm gonna continue with `-dbhelp` since description is changed and dbms can be used for sorting.

npetzall:
I've did some simple update to the `-dbhelp`
image

npetzall:
All I have left is update the build instructions.

npetzall:
I'll have it fixed tomorrow. ConfigIT broke, Config.dumpUsage needs an update to output gets properties().getProperty("dbms")

PR #1006, opened 2023-02-23
Fix readme example

jesperolsson-se:


npetzall:
Not sure if schema needs to be set to public or if it's default resolved to public.

jesperolsson-se:
> Not sure if schema needs to be set to public or if it's default resolved to public.

It works without specifying. I guess we have a trade-off between shorter command and better explanation. If it matters, `--schema` is also parsed correctly.

If you want to try out quickly:
```
docker run --name some-postgres -p 5432:5432 -e POSTGRES_PASSWORD=mysecretpassword -d postgres
psql -d postgres -U postgres -h localhost
```
```
CREATE TABLE foo (col int);
...
\q
```
```
java -jar schemaspy-6.1.0.jar -t pgsql11 -dp ~/Downloads/jdbc-driver.jar -db postgres -host localhost -port 5432 -u postgres -p mysecretpassword -o DIRECTORY
```

npetzall:
I've promised to verify a thing on 15 so I'm gonna set it up, just to check. I'm also gonna check with a custom schema. Most to see if public is default set by driver or not.

jesperolsson-se:
I would prefer that this PR is not delayed since it affects newcommers in a big way.
The entire `--schema` discussion is already a bit off-topic. If you prefer, I can change to `--schema public`

npetzall:
I can accept specifying public directly.

Since the only bad situation is when they do the demo without and they want to try with their own and it has a different schema. 

We might have some logging for the situation where public might be missing but not public exists but the need to use a different schema.

npetzall:
I created an issue and we handle it maybe as a later improvement.

PR #1010, opened 2023-02-28
971 dbhelp

npetzall:
image


src/main/resources/org/schemaspy/types/sqlite.properties
```diff
@@ -2,7 +2,7 @@
 # see http://schemaspy.org/dbtypes.html
 # for configuration / customization details
 #
-
+dbms=SQLite
 description=SQLite
```
jesperolsson-se, line 6: Standard?

jesperolsson-se (approved):

src/main/resources/org/schemaspy/types/sqlite.properties
npetzall: No, there are two different drivers like in the jTDS and Microsoft case, so this is the SQLite and the other one is xerial, the xerial is better so I don't want to call this standard.

PR #1035, opened 2023-03-06
chore(docs): explain -maxdet

### LINKED ISSUE #957: columns
## Expected Behavior
Running the Docker version of Schemaspy against SQLServer databases.

Some databases the Relationship diagram has the table  with COLUMN information it

On another the COLUMN information isn't displayed.

1.
2.
3.
4.

## Context

* Database type (e.g. MySQL):
* Database driver version (is you use a special one):

## Your Environment

* Version used:
* Environment name and version (e.g. Chrome 39):
* Operating System and version (e.g. Windows 10):
* Java version:

npetzall:
fixes #957

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -99,7 +99,7 @@ Html report related
     Path to custom mustache template/css directory, needs to contain full set of templates.
     Bundled templates can be found in jar '/layout' and can be extracted with jar tool or any zip capable tool.
 [-maxdet number]
-    Limit for when tables should be shown with details.
+    Threshold for details in diagrams, increase if you still want details, checked against sum of tables and views.
 [-css fileName]
     Use a custom stylesheet. Bundled stylesheet can be extracted from jar(using zip capable tool), path '/layout/schemaSpy.css'
 [-desc description]


docs/source/configuration/commandline.rst
```diff
@@ -99,7 +99,7 @@ Html report related
     Path to custom mustache template/css directory, needs to contain full set of templates.
     Bundled templates can be found in jar '/layout' and can be extracted with jar tool or any zip capable tool.
 [-maxdet number]
-    Limit for when tables should be shown with details.
+    Threshold for details in diagrams, increase if you still want details, checked against sum of tables and views.
```
jesperolsson-se, line 102: Make more concise.
Maybe something like "Maximum number of columns displayed in the entity-relationship diagram"

Possibly add default value and default behaviour

npetzall: Since your example is incorrect, there's something wrong with the text or the "option"

Show table/view details(all columns) as long as the sum(tables + views) are lower than this assigned value. Default is 400 i think.

jesperolsson-se: Yeah, this is definitely not obvious at all. In addition to the question of what `details` is, what `diagrams` are we talking about? Compact, large, one degree, two degrees?

I think I need a better understanding of the scenario where we use this, before I can suggest improvements.

npetzall: https://github.com/schemaspy/schemaspy/blob/07f5f074c908a68b598848e513693b487d3e3636/src/main/java/org/schemaspy/SchemaAnalyzer.java#L358

npetzall:
Hmm, i think the one I changed should be removed. It's related to diagrams not directly to html report

PR #1112, opened 2023-03-26
Add Maven badge/shield

marcelstoer:
Also moved the Docker badge to the front of the list. I find badges that lead to downloadable artifacts the most important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,12 @@
 [![Logotype](https://github.com/schemaspy/schemaspy/blob/master/docs/source/_static/schemaspy_logo.png)](https://schemaspy.org/)
 
+[![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.schemaspy/schemaspy/badge.svg)](https://maven-badges.herokuapp.com/maven-central/org.schemaspy/schemaspy/)
+[![Docker Pulls](https://img.shields.io/docker/pulls/schemaspy/schemaspy.svg)](https://hub.docker.com/r/schemaspy/schemaspy/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,12 @@
 [![Logotype](https://github.com/schemaspy/schemaspy/blob/master/docs/source/_static/schemaspy_logo.png)](https://schemaspy.org/)
 
+[![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.schemaspy/schemaspy/badge.svg)](https://maven-badges.herokuapp.com/maven-central/org.schemaspy/schemaspy/)
```
jesperolsson-se, line 3: How come you went with this heroku app, instead of [shields.io](https://shields.io/)?

jesperolsson-se (approved):

jesperolsson-se:
@marcelstoer
Good point! We should probably go one step further and group them in different lines. For example, one line for badges about downloadable artifacts and one line for badges about code quality. 

Would you like to make a proposal, in a new PR?

README.md
marcelstoer: shields.io is my go-to badges generator. However, AFAIK they don't have Maven Central badges with automatic redirect (i.e. only the badge, not the Central link).

marcelstoer:
@jesperolsson-se 

I personally prefer a single line of badges - use it like that on all my projects. I am sure there are valid arguments to be found for various types of arrangements. The maintainers should structure the badges according to their personal preferences. I'm not inclined to intervene 😄 

With somewhat strict grouping you'd end up with this.

Screen Shot 2023-03-26 at 22 15 44

Issue #1068, opened 2023-03-13
Can't run integration tests from IDE


npetzall:
It's not possible to run Integration tests from IDE due to a requirement that software is packaged as Jar

If there is a workaround let's document it, if not we should consider brining some "dead code" back, by handling the case when layout.url() is file protocol instead of jar.

npetzall:
https://www.jetbrains.com/help/idea/work-with-tests-in-maven.html

From intellij, it's a complex workaround. Strange that it's not autogenerated the same way it's done with gradle.

npetzall:
There is an option to create a ZipFileSystem, which means that we might be able to treat the sources as Path objects, and copy them to target. More or less, if jar create ZipFileSystem and then walking the Path and copy.

More the same handling for both, just that the ZipFileSystem needs to be setup.

PR #1113, opened 2023-03-26
#1068/integration tests

jesperolsson-se:


npetzall:
We can go with this for now, names are a bit strange. JarCopy sounds like you are copying a Jar and UrlCopy sounds like you're copying a URL and not actually contents hidden beneath the Url.

npetzall (approved):

jesperolsson-se:
I agree. Does either `CopyFromJar` or `CpJar` sound better to you?

PR #1123, opened 2023-03-31
Configurable flexmark options

akruis:
Hi,

we use schemaspy to document an internal database project. During the development it is very convenient to use the markdown editor of our Gitlab instance to write documentation for database objects using Gitlab-flavored markdown. Unfortunately schemaspy uses pegdown flavored markdown.

The flexmark processor is very flexible and can render many different markdown variants, but currently its configuration is hard wired. This merge request is a prove of concept to make the flexmark configuration an runtime option.

My Java is not the best any more and I'm not familiar with the implementation details of the schemaspy configuration. Therefore I used system properties instead of a proper integration with the schemaspy configuration. Clearly not production quality, but better than nothing.

CU Anselm

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,12 +80,12 @@
             <dependency>
                 <groupId>com.vladsch.flexmark</groupId>
                 <artifactId>flexmark</artifactId>
-                <version>0.34.32</version>
+                <version>0.64.0</version>
             </dependency>
             <dependency>
                 <groupId>com.vladsch.flexmark</groupId>
                 <artifactId>flexmark-profile-pegdown</artifactId>
-                <version>0.34.32</version>
+                <version>0.64.0</version>
             </dependency>
             <dependency>
                 <groupId>org.json</groupId>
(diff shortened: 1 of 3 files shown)


npetzall:
Sorry for the late reply I've been out sick, I'll try to have a look at how the configuration is done and what's required to configure flexmark-java.

I'm hoping that there might be a simpler solution than having to go thru JavaScript or at least make it simpler for users.

npetzall:
From flexmark-java:
> For Versions 0.62.2 or below, Java 8 or above, Java 9+ compatible. For Versions 0.64.0 or above, Java 11 or above.

Might need to use <  0.64.0 for now, or make a decision on bumping java requirement. But that's a totally different issue.

akruis:
> Sorry for the late reply I've been out sick, I'll try to have a look at how the configuration is done and what's required to configure flexmark-java.
> 
> I'm hoping that there might be a simpler solution than having to go thru JavaScript or at least make it simpler for users.

Sure. Once you have a suitable set of flexmark options, you could provide a ready-to-use configuration. But being able to change the configuration details an runtime is valuable too.

> Might need to use < 0.64.0 for now, or make a decision on bumping java requirement. But that's a totally different issue.
I upgraded, because I first thought that my rendering problems might be caused by a flexmark bug. And then I was to lazy to look for the documentation of older flexmark versions. Version 0.62.2 is the last version, that supports Java 8.

npetzall:
@akruis not ready to use but a simpler configuration file, so that it's easy to modify and use your own configuration.

PR #1132, opened 2023-04-04
refactor(orderer): Replace while-iterators

jesperolsson-se:


npetzall:
Why?

npetzall:
Okay, I thought the whole reason for the iterator was to allow safe modifications to a collection.

PR #1177, opened 2023-05-26
Add note about JAR types

### LINKED ISSUE #1174: "no main manifest attribute" when running from JAR
Running SchemaSpy 6.2.2 from the JAR on Maven Central I get this (quite rightly so)

> no main manifest attribute, in /root/.m2/repository/org/schemaspy/schemaspy/6.2.2/schemaspy-6.2.2.jar

This is mentioned at https://github.com/schemaspy/schemaspy/issues/157#issuecomment-1483934495 and I simply forget to use the JAR with the `app` classifier.

The issue here serves two purposes:
- documentation for those running into the same issue and searching the issue list on GitHub
- pending item for myself as I will submit a PR that adds a note about this to README

marcelstoer:
Fixes #1174

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,23 +8,34 @@
 [![Coverage](https://sonarcloud.io/api/project_badges/measure?project=org.schemaspy%3Aschemaspy&metric=coverage)](https://sonarcloud.io/dashboard?id=org.schemaspy%3Aschemaspy)
 [![Gitter](https://badges.gitter.im/schemaspy/schemaspy.svg)](https://gitter.im/schemaspy/schemaspy?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
-**SchemaSpy** is a database metadata analyzer. It helps your database administors and developers visualize, navigate and understand your data model. With an easy-to-use HTML-based report, traversing the entity-relationship diagram has never been simpler. Our product showcase is available at http://schemaspy.org/sample/index.html.
+**SchemaSpy** is a database metadata analyzer. It helps your database administrators and developers visualize, navigate and understand your data model. With an easy-to-use HTML-based report, traversing the entity-relationship diagram has never been simpler. Our product showcase is available at http://schemaspy.org/sample/index.html.
 
 [![SchemaSpy sample](http://schemaspy.org/img/example_page_epivirusurf.png)](http://schemaspy.org/samples/epivirusurf)
 
 # Installation
 
-SchemaSpy is a stand alone tool. Just download the lastest
-[JAR file](https://github.com/schemaspy/schemaspy/releases/download/v6.2.4/schemaspy-6.2.4.jar)
-or [Docker image](https://hub.docker.com/r/schemaspy/schemaspy/) and you're
-ready to go!
+SchemaSpy is a standalone application without GUI.
+Just download the latest [JAR file](https://github.com/schemaspy/schemaspy/releases/latest) or [Docker image](https://hub.docker.com/r/schemaspy/schemaspy/) and you're ready to go!
+To use SchemaSpy from Maven, please see the [Maven chapter](#maven) below.
 
 ```
+# replace '6.2.4' with latest version
 curl -L https://github.com/schemaspy/schemaspy/releases/download/v6.2.4/schemaspy-6.2.4.jar \
     --output ~/Downloads/schemaspy.jar
 ```
 
 > For unreleased bug fixes and features-in-progress, download our [snapshot JAR](https://schemaspy.org/schemaspy/download.html) or use Docker tag `snapshot`
+
+## Maven
+
+SchemaSpy releases two types of JAR files: a bare-bone JAR and a fat JAR including all dependencies.
+Both JARs are published to Maven Central.
+The fat JAR is also attached to releases on GitHub.
+The "maven central" badge at the top of this page will take you straight to the latest version on Maven Central.
+
+The Maven GAV of the two artifacts is as follows:
+- bare-bone JAR: `org.schemaspy:schemaspy:<version>`
+- fat JAR: `org.schemaspy:schemaspy:<version>:app` **← note the `app` classifier**
 
 # Quick start
 
@@ -70,7 +81,7 @@ Be sure to check out the guides provided by the community later in this README.
 
 ## On-demand database documentation
 
-The prefered way to document databases is through entity-relationship (ER) diagrams.
+The preferred way to document databases is through entity-relationship (ER) diagrams.
 However, drawing these diagrams manually is such a time-consuming and error-prone
 process that we hardly ever draw them in practice. When the diagrams *are* drawn,
 they rarely stay up-to-date. With SchemaSpy, this is no longer a problem.
@@ -208,17 +219,17 @@ The BibTeX entry for LaTeX users is:
 SchemaSpy is built using maven and we utilize the maven wrapper.  
 __Windows__ `mvnw.cmd package`    
 __Linux__ `./mvnw package`  
-The resulting application can be found in `target`  
+The resulting application can be found in `target`
 
 ### Analyzing
 You need your own SonarQube:  
 https://hub.docker.com/_/sonarqube/  
 __Windows__ `mvnw.cmd -P sonar clean verify -Dsonar.host.url=http://$(boot2docker ip):9000 -Dsonar.jdbc.url="jdbc:h2:tcp://$(boot2docker ip)/sonar"`  
-__Linux__ `./mvnw -P sonar clean verify`  
+__Linux__ `./mvnw -P sonar clean verify`
 
 Watch results at:  
 __Linux__ `http://localhost:9000`  
-__Windows__ `http://$(boot2docker ip):9000`  
+__Windows__ `http://$(boot2docker ip):9000`
 
 ## Documentation
 Built using Python  


marcelstoer:
Are there any objections to this PR marked as "important"?

marcelstoer:
Can I propose to merge or close this one?

README.md
```diff
@@ -8,24 +8,35 @@
...
 ```
-curl -L https://github.com/schemaspy/schemaspy/releases/download/v6.2.4/schemaspy-6.2.4.jar \
+# replace '6.2.4' with latest version
+curl -L https://github.com/schemaspy/schemaspy/releases/download/v6.2.2/schemaspy-6.2.4.jar \
     --output ~/Downloads/schemaspy.jar
 ```
 
 > For unreleased bug fixes and features-in-progress, download our [snapshot JAR](https://schemaspy.org/schemaspy/download.html) or use Docker tag `snapshot`
 
+## JAR files and dependencies
```
npetzall, line 28: Rename this section to only maven and short description of the two published jar-files and their differences as normal jar and fat-jar(more batteries included, not all but more)

npetzall (commented):
@jesperolsson-se can you have a look since you've been most opinionated regarding the readme.

We want it to be simple, and I would say maven is currently advanced usage.

README.md
```diff
@@ -8,24 +8,35 @@
...
 [![SchemaSpy sample](http://schemaspy.org/img/example_page_epivirusurf.png)](http://schemaspy.org/samples/epivirusurf)
 
 # Installation
 
-SchemaSpy is a stand alone tool. Just download the lastest
-[JAR file](https://github.com/schemaspy/schemaspy/releases/download/v6.2.4/schemaspy-6.2.4.jar)
-or [Docker image](https://hub.docker.com/r/schemaspy/schemaspy/) and you're
-ready to go!
+SchemaSpy is a standalone application without GUI. 
+Just download the latest [JAR files](#jar-files-and-dependencies)(see below) or [Docker image](https://hub.docker.com/r/schemaspy/schemaspy/) and you're ready to go!
```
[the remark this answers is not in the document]
marcelstoer: Thanks, I can certainly change the PR according to your feedback.

The "latest" URL https://github.com/schemaspy/schemaspy/releases/latest takes you to the latest release but you cannor statically deep-link the actual _asset_ to download as its name is dynamic. Hence, you will continue to have to update the README for every release.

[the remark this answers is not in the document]
npetzall: Yeah I know, the example would need to be updated. But clicking the link would get them to the correct page to download.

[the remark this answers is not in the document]
npetzall: But I would hold any changes until @jesperolsson-se has had a look. He might be of a different opinion.

jesperolsson-se:
The problem doesn't seem to be that the README is unclear, but rather with Maven Central. Perhaps the issue can be resolved there, at the source, as Nils suggested?

Regarding the changes to the README, I agree with Nils. Maven is advanced use and should not be discussed until later in the README (if at all), and we want to avoid telling the reader to jump back and forth between sections.

The discussion about `latest` is good. I can't really check right now, but removing the version number from the jar name sounds completely reasonable.

marcelstoer:
I appreciate your feedback Jesper, thanks - it appears I disagree completely 😄 

Here's my perspective.

I ran into the issue described at #1174. Finding the hint to a solution in an issue comment is far from ideal I argue. If my project is publishing multiple assets and/or publishes those to multiple locations, I should want to document this for my users. Apparently it _is_ important as picking the wrong asset for the wrong use case yields errors. The first place I always look is the project's README.

> The problem doesn't seem to be that the README is unclear, but rather with Maven Central. Perhaps the issue can be resolved there, at the source

1. It's not that the current documentation is unclear, the fact that your publishing multiple assets is simply missing. IMO there's nothing wrong on the Maven Central side or your use thereof.
2. I wouldn't know how to solve this on the Maven end unless you abandon the two-artifacts strategy.

> Maven is advanced use and should not be discussed until later in the README (if at all)

1. I see Maven as one two prevalent Java(-universe) build and packaging tools and it's soon approaching dinosaur-status. Maven Central is _the_ place to get artifacts from for most of us. Calling Maven "advanced use" is not something I can relate to at all.
2. If you are publishing to Central - which I am very happy about - I feel the README should absolutely talk about it, even more so if Maven users should know about a special classifier!

> and we want to avoid telling the reader to jump back and forth between sections.

That's where we agree! Hence, I proposed this one paragraph with no back links (i.e. no back and forth) that describes the different artifacts and how to get them. I feel it is good UX to cross link paragraphs if they are organized in a logical directed way.

Again, I'm fine with you closing this PR if my arguments are not convincing enough.

npetzall:
If we build some context. SchemaSpy isn't yet a library. So it's more of an application.

There will be a transition period I would say before it can be used as a library and a dependency in that matter. That's the reason for the usage of classifier and dual artifacts.

If we used the classifier in the name under release it would be more obviouse that the classifier is used.

As for the publishing to maven, it's more or less for the usage as a dependency which is way more advanced than downloading from release and using. Since it doesn't involve Java knowledge and a build system. It's even advanced since it can't yet be used as a library. The closes thing is to run it with maven:exec or Gradle equivalent, where it's just download and execute.

So from the top of my head.
We have a maven section since there is that use-case.

The initial just points towards releases, the version is bundled inside the jar, so in releases we could probably remove the version number and possibly maybe do linking thru `latest` url. Need to look into that.

The maven badge, is linked to the maven section with the explanation and the actual link to maven central.

As for the future, we could publish as two separate artifacts, might be more artifacts if we actually modularized SchemaSpy.

npetzall:
@marcelstoer i have this feeling but haven had time to test.

If you add SchemaSpy with the `app` classifier as a dependency will it still include all dependencies specified in the pom. Since I don't think classifiers have their own pom.

If that is the case, the fat-jar needs to be it's own artifact with a reduced pom.

marcelstoer:
@npetzall I updated the README as per your comments. Regardless of how you will proceed with or without changing the multi-artifact publishing strategy I feel this change adds useful clarifications for some while not impeding whose you don't need them.

README.md
```diff
@@ -8,24 +8,35 @@
...
 [![SchemaSpy sample](http://schemaspy.org/img/example_page_epivirusurf.png)](http://schemaspy.org/samples/epivirusurf)
 
 # Installation
 
-SchemaSpy is a stand alone tool. Just download the lastest
-[JAR file](https://github.com/schemaspy/schemaspy/releases/download/v6.2.4/schemaspy-6.2.4.jar)
-or [Docker image](https://hub.docker.com/r/schemaspy/schemaspy/) and you're
-ready to go!
+SchemaSpy is a standalone application without GUI. 
+Just download the latest [JAR file](#jar-files-and-dependencies)(see below) or [Docker image](https://hub.docker.com/r/schemaspy/schemaspy/) and you're ready to go!
```
npetzall, line 18: I would want 
`Just download the latest [JAR file](https://github.com/schemaspy/schemaspy/releases/atest) or [Docker image]`

The link doesn't work since the header changed. But for non-java people we want to make it as simple as possible and just link to release page.

README.md
```diff
@@ -8,24 +8,35 @@
...
+curl -L https://github.com/schemaspy/schemaspy/releases/download/v6.2.2/schemaspy-6.2.4.jar \
     --output ~/Downloads/schemaspy.jar
 ```
 
 > For unreleased bug fixes and features-in-progress, download our [snapshot JAR](https://schemaspy.org/schemaspy/download.html) or use Docker tag `snapshot`
 
+## Maven
+
+SchemaSpy releases two types of JAR files: a bare-bone JAR and a fat JAR including all dependencies.
+Both JARs are published to Maven Central alongside the standard Javadoc and sources JARs.
```
npetzall, line 31: Remove this line, we assume that people familiar with maven knows that this is required for publishing.

README.md
```diff
@@ -8,24 +8,35 @@
...
+In addition, the fat JAR is attached to releases on GitHub.
+Navigate to the [releases page](https://github.com/schemaspy/schemaspy/releases), pick a version and look for the attached `schemaspy-<version>.jar` asset.
```
npetzall, line 33: Remove this line, if they are actually reading about Maven and for maven this information isn't important.

README.md
```diff
@@ -8,24 +8,35 @@
...
+
+As for Maven, the "maven central" badge at the top of this page takes you straight to the latest version on Maven Central.
```
npetzall, line 35: replace line with 

"maven central" badge at the top of this page will take you straight to the latest version on Maven Central.

README.md
```diff
@@ -8,24 +8,35 @@
...
+The Maven coordinates of the artifacts are as follows:
+- bare-bone JAR: `org.schemaspy:schemaspy`
```
npetzall, line 37: add version to have all parts of GAV

- bare-bone JAR: `org.schemaspy:schemaspy:<version>`

npetzall (changes requested):
Minor fixes and reduction of text.
But you are correct, readme can be updated again when a "strategy" has been put in place.

npetzall (approved):

marcelstoer:
I now see where you want to take this. Reshaping that chapter to focus solely on Maven required IMO a few minor tweaks to make this a consistent read.

Issue #487, opened 2019-01-02
Add view for Custom/User-Defined Types (UDTs) 


zam6ak:
I am sure all DBs could benefit by having a view that shows UDTs.
This has been mentioned before in #389, #395

The view could show type name, definition, comments, and related constraints.
As an example, here is a recap of an example I gave for PostgreSQL with demo database (see below for details)

* DB has 1 **domain**, (`public.year`) which is not shown anywhere 
  * the column `release_year `in table `film `does show correct definition of `year`
* DB has 1 **enum** type (`public.mpaa_rating`) which is not shown anywhere
  * the column `rating `in table `film `does show correct definition of `mpaa_rating`

---
Tested using:

* DB:        PostgreSQL v11
* Driver:    [PostgreSQL JDBC 4.2 Driver, 42.2.5](https://jdbc.postgresql.org/download/postgresql-42.2.5.jar)
* Schemaspy: [schemaspy-6.1.0-SNAPSHOT](https://github.com/schemaspy/schemaspy/raw/gh-pages/schemaspy-6.1.0-SNAPSHOT.jar)
* Sample DB: [DVD Rental](http://www.postgresqltutorial.com/postgresql-sample-database/)
* Java ver:  1.8

dopeddude:
+1

albrechtd:
+1 – PostgreSQL UDT support would be mega!

OrestisNer:
Hey @npetzall. UDT support would be awesome feature for the project. 
Do you think that there is a roadmap for this?
Also, I would like to contribute if you guide me a bit :)

npetzall:
@OrestisNer help would be much appreciated. In our readme.md we have a link to gitter, you can joins us there and I'll try to be as available as possible. Seems we are in the same time zone, that's good.

Spontaneous thoughts are:
1, Accessing and retrieving information from database.
2. Populate the internal database model with the retrieved data.
4. Add information to xml report.
3. Add information to html report.

Where 1 will impact all other steps. So I think we should investigate a cpl of databases so that we can create  a common model for them all. Less might be more.

So adding comments here for UDT information that can be extracted from
* postgresql
* mysql
* mssql
* oracle
* informix

The more we can use JDBC the better, but having the option for custom SQL is probably going to be a requirement.

andreievg:
For now, can do manual introspection of postgres schema and add to column comment:
[...]

PR #1182, opened 2023-06-01
Adding a tab for the types 

samdus:
Our research group has an extensive use of domains and types. Therefore we would love to have a tab that describes them. This PR proposes a way to do this, along with an implementation for PostgreSQL. We will provide queries for other dbms over the summer.

Here's a preview of how the different PostgreSQL types are printed (Type of types are dynamics since each dbms has its own definitions): 

image

### DIFF
--- a/src/test/java/org/schemaspy/integrationtesting/pgsql/PgSqlTypesIT.java
+++ b/src/test/java/org/schemaspy/integrationtesting/pgsql/PgSqlTypesIT.java
@@ -65,7 +65,7 @@ public class PgSqlTypesIT {
         String[] args = {
             "-t", "pgsql",
             "-db", "test",
-            "-s", "public",
+            "-s", "type_tests",
             "-cat", "%",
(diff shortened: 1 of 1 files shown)


npetzall:
Would this be solving #487?

VietThan:
@samdus @npetzall not sure if this is proper etiquette but I'm interested in this feature being merged in.

npetzall:
@VietThan add a thumbs up to the description and we'll know that more people are interested in it.

I've finished up another about giving the ability to use AsciiDoc so I'm looking at multi arch for Docker and after that I'll have a look at this.

Just need to make sure that it's generic enough so that other Dbms than Pgsql can use it without becoming too strange.

npetzall:
So mysql doesn't have UDT. MSSQL have both CRL and Table variants for it. Need to test them. But I think they both will fit. Still need to look a bit at Oracle and Informix. But since I'm no DB in any DB I'm mostly guessing.

PR #1189, opened 2023-06-13
feature(Asciidoc): Allow the user to use Asciidoc as a markup language

### LINKED ISSUE #938: Markdown creates invalid links
*Given*
Comment in Markdown with a reference link to a non-existing table.

*When*
Analyzing with Html enabled.

*Then*
Comment has a reference link with path containing `null`

**Expected**
That it wouldn't be processed by Schemaspy and original data is presented.

samdus:
We are currently using Asciidoc rather than Markdown to document our databases. This PR adds a new parameter (-asciidoc) that replaces the usages of the Markdown class with the Asciidoc class.

### Implementation details
* Change the pages hashmap singleton with a singleton that encloses the complete MarkupProcessor
* Defaults to Markdown
* Adds a -asciidoc parameter
* Changes the way the reference links are handled to uniformize markdown processing with asciidoc
** Rather than adding `[link]: path` at the end of the Markdown string, it replaces the `[link]` directly.

### Bugfixes
* It also fixes #938

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -106,6 +106,8 @@ Html report related
     Use a custom stylesheet. Bundled stylesheet can be extracted from jar(using zip capable tool), path '/layout/schemaSpy.css'
 [-desc description]
     Add a description to the index page.
+[-asciidoc]
+    Uses asciidoc rather than markdown when processing descriptions
 
 DataTables related
 ------------------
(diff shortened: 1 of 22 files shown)


ja11sop:
Just adding my support for this PR - we also use Asciidoc for all our DB documentation and this would be fantastic to see.

npetzall:
I'm going to start looking into this. I might pull your commits into a new PR. So you don't need to update it. I'll take care of it.

PR #1205, opened 2023-06-27
chore(classpath): Remove obsolete use case

jesperolsson-se:


npetzall (approved):
Yeah, this doesn't work, the use-case might have been nested jars but that doesn't work. Getting something from what probably would be the delegating classloader of urlclassloader and add it to urlclassloader is just unnecessary. So this can be removed.

PR #1229, opened 2023-07-24
chore!: drop support for -loadjars

jesperolsson-se:


npetzall:
Since the parameter setup is removed, could you do the description key at the same time else it might get forgotten.

npetzall (approved):
Finally we can start to remove this.

Issue #1031, opened 2023-03-06
(CLI): Long format is inconsistent


npetzall:
Example

`-t` which has a long format of `--database-type`

and then we have 

`-dp` which has a long format of `-driverPath`

We should settle on a singe format and not mix.

npetzall:
There should be consistency between long args, but the should also be in synch with short args.

Everything doesn't need a short args.

PR #1251, opened 2023-08-05
#763 orphans

### LINKED ISSUE #763: Full schema diagram including relationships and orphan tables
This branched off from #376 where @vlehm, @AndreKR and I seem to have requested the same feature.

## Expected Behavior
SchemaSpy should generate a complete visual representation of "the schema" (a vague term).

## Current Behavior
SchemaSpy generates one diagram showing relationships and a dedicated PNG/SVG file for every orphan table.

## Possible Solution
A) Include all orphan tables in the "Relationships" diagram. Easy to reason about as even orphan tables somehow fall under the relationships category as there are "_no_ relationships" to such tables. Maybe give users an option to turn this behavior off (enabled by default).
B) Keep "Relationships" and "Orphan Tables" as is but introduce a new aggregated diagram. I could live with that but I see little value in that.

## Context
We are automatically generating SchemaSpy diagrams in the build pipeline to include in the @antora documentation for several dozen projects: create an H2 file-based database, run Liquibase changeset against it, run SchemaSpy against it, copy SchemaSpy diagram files to documentation folder, Git-commit changes.

jesperolsson-se:
Fixes #763

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -91,6 +91,8 @@ Html report related
     Skip generation of html report.
 [-noimplied]
     Don't look for implied relationships.
+[--no-orphans]
+    Don't include orphan tables in the relationships graph.
 [-nopages]
     Just list data as one long list instead of pages.
 [-rails]
--- a/src/main/java/org/schemaspy/cli/CommandLineArguments.java
+++ b/src/main/java/org/schemaspy/cli/CommandLineArguments.java
@@ -104,6 +104,15 @@ public class CommandLineArguments {
 
     @Parameter(
         names = {
+            "--no-orphans",
+            "schemaspy.no-orphans"
+        },
+        descriptionKey = "noorphans"
+    )
+    private boolean noOrphans = false;
+
+    @Parameter(
+        names = {
             "-rails",
             "schemaspy.rails"
         },
@@ -491,6 +500,8 @@ public class CommandLineArguments {
     public boolean withImpliedRelationships() {
         return !noImplied;
     }
+
+    public boolean withOrphans() { return !noOrphans; }
 
     public boolean isRailsEnabled() {
         return railsEnabled;
--- a/src/main/resources/commandlinearguments.properties
+++ b/src/main/resources/commandlinearguments.properties
@@ -5,6 +5,7 @@ license=Print license, it will first print GPL and then LGPL (LGPL is addition t
 debug=Enable debug logging
 nohtml=Skip html generation
 noimplied=Don't look for implied relationships
+noorphans=Don't include orphan tables in the relationships graph
 rails=Use Rails-based naming convention to find relationships between logical foreign keys and primary keys.
 meta=Single schema analysis file path to SchemaMeta-xml, when running '-all' or '-schemas' path to directory containing SchemaMeta-xmls with pattern (DatabaseName|Schema).meta.xml
 sso=Remove requirement for user
--- a/src/main/java/org/schemaspy/SchemaAnalyzer.java
+++ b/src/main/java/org/schemaspy/SchemaAnalyzer.java
@@ -382,7 +382,7 @@ public class SchemaAnalyzer {
             isOneOfMultipleSchemas
         );
 
-        DotFormatter dotProducer = new DotFormatter(runtimeDotConfig);
+        DotFormatter dotProducer = new DotFormatter(runtimeDotConfig, commandLineArguments.withOrphans());
 
         File diagramDir = new File(outputDir, "diagrams");
         diagramDir.mkdirs();
--- a/src/main/java/org/schemaspy/cli/CombinedDefaultProvider.java
+++ b/src/main/java/org/schemaspy/cli/CombinedDefaultProvider.java
@@ -14,6 +14,7 @@ public class CombinedDefaultProvider implements IDefaultProvider {
         "schemaspy.debug",
         "schemaspy.nohtml",
         "schemaspy.noimplied",
+        "schemaspy.no-orphans",
         "schemaspy.vizjs",
         "schemaspy.all",
         "schemaspy.rails",
--- a/src/main/java/org/schemaspy/output/dot/schemaspy/DotFormatter.java
+++ b/src/main/java/org/schemaspy/output/dot/schemaspy/DotFormatter.java
@@ -43,9 +43,9 @@ public class DotFormatter {
 
     private final DotSummaryFormatter dotSummaryFormatter;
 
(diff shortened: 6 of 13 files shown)


src/main/java/org/schemaspy/cli/CommandLineArguments.java
```diff
@@ -102,6 +102,15 @@ public class CommandLineArguments {
     )
     private boolean noImplied = false;
 
+    @Parameter(
+            names = {
+                    "--noorphans",
```
npetzall, line 107: Should this be `-noorphans` and `--no-orphans`

src/main/java/org/schemaspy/cli/CommandLineArguments.java
```diff
@@ -102,6 +102,15 @@ public class CommandLineArguments {
     )
     private boolean noImplied = false;
 
+    @Parameter(
```
npetzall, line 105: Is there a reason why this is added to CommandLineArguments instead of DotConfig?

npetzall (commented):
Unsure about the placement of the argument, I would have put it in DotConfig. But there might exist a good reason why not.

src/main/java/org/schemaspy/cli/CommandLineArguments.java
jesperolsson-se: I couldn't find any SchemaSpy standard for this, and existing parameters are inconsistent. I'm happy to change if there's some thought behind it that I've missed.

> `-noorphans`

In my mind, single hyphens exist because of convenience: Allowing commonly used parameters to be expressed with much less typing. So they should be single letters. In other words, not every parameter needs a short version.

What purpose would long options with single hyphens serve?

> `--no-orphans`

Most of SchemaSpy's multi-world parameters do _not_ use kebab case (e.g., `--dbhelp` and `--outputDirectory`) but some do (e.g., `--no-html` and `--no-implied`). Personally, I prefer the kebab version and can update if you want.

npetzall: There is an issue for the inconsistency with the command line args. Should we say kebab and then do breaking changes to make it consistent.

Same goes for the short syntax, `-no` would be no orphans if it doesn't collide with another argument.

src/main/java/org/schemaspy/cli/CommandLineArguments.java
jesperolsson-se: Orphans is a SchemaSpy concept, not a `dot` one, similar to how `--no-implied` is in `CommandLineArguments` while `-rankdirbug` ([link](https://graphviz.org/docs/attrs/rankdir/)) is in `DotConfig`.

Expressed differently, the first relates to our algorithms for what parts of the database we should display while the second is how the graph itself should be rendered/presented.

src/main/java/org/schemaspy/cli/CommandLineArguments.java
jesperolsson-se: Sounds good for moving forward with this PR. Let's create a new issue for consistency/policy.

npetzall: There is an issue

src/main/java/org/schemaspy/cli/CommandLineArguments.java
npetzall: We have Dot concept in SchemaSpy and the configuration is for that.

Configuration would be added to `org.schemaspy.output.dot.DotConfig` it would be used in `org.schemaspy.output.dot.schemaspy.DotFormatter`

src/main/java/org/schemaspy/cli/CommandLineArguments.java
npetzall: #1031

src/main/java/org/schemaspy/cli/CommandLineArguments.java
jesperolsson-se: I've created issue #1252 to continue the discussion. How about we don't let this PR grow stale waiting for a decision? Even in the worst case, it's better to merge and then submit a new change.

npetzall: Yeah, sure go ahead.

My thinking is that both `no implied` and `rankdir` both control content of dot syntax file and nothing else in the system. (Might be wrong about no implied. But for this setting it only effects dot file content it doesn't remove orphans from the complete system. So over exposing it or lack of context, it's global and you need to look up the resource key or check other usages to find out.

I'm unsure about no implied since it's in this file.

src/main/java/org/schemaspy/cli/CommandLineArguments.java
```diff
@@ -102,6 +102,15 @@ public class CommandLineArguments {
...
+        names = {
+            "-no", "--no-orphans",
```
npetzall, line 107: Should we really have `-no` and `schemaspy.no`, if so, the latter needs to be added to CombinedDefaultsProvider.

src/main/resources/commandlinearguments.properties
```diff
@@ -5,6 +5,7 @@ license=Print license, it will first print GPL and then LGPL (LGPL is addition t
 debug=Enable debug logging
 nohtml=Skip html generation
 noimplied=Don't look for implied relationships
+noorphans=Don't include orphan tables in the main graph
```
npetzall, line 8: What is the main graph?

npetzall (changes requested):
If `schemaspy.no` should be available it should be added to CombinedDefaultsProvider.

src/main/java/org/schemaspy/cli/CommandLineArguments.java
npetzall: I'm fine either way. It could be handled in #1031 if we just expand it to include consistency for alla args and not just long ones. Since I think it's probably best to check both so they are in synch. For now it's mainly that it's incomplete and I'm fine going either way.

docs/source/configuration/commandline.rst
```diff
@@ -91,6 +91,8 @@ Html report related
     Skip generation of html report.
 [-noimplied]
     Don't look for implied relationships.
+[-no]
```
npetzall, line 94: The option is `--no-orphans`

npetzall (changes requested):
Args documentation has wrong arg.

docs/source/configuration/commandline.rst
jesperolsson-se: Nice catch! The comment was also wrong ("main graph").

npetzall (approved):

PR #1255, opened 2023-08-11
Added cli parameter -hidelabel

rahulpr27git:
Added cli parameter `-hidelabel` to hide the default or any label on the diagram

### DIFF
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -230,5 +230,5 @@ Diagram related
     Use embedded viz.js instead of Graphviz. Useful when graphviz isn't installed. Memory is set to 64 MB, if you receive ther error "Cannot enlarge memory arrays" please report this to us.
 [-degree 1 or 2]
     Limit the degree of separation (1 shows less, 2 is default), 1 is a good option for large databases with lots of relationships.
-[-hidelabel]
+[-hideLabel]
     Hide the default label.
--- a/docs/source/configuration/commandline.rst
+++ b/docs/source/configuration/commandline.rst
@@ -230,3 +230,5 @@ Diagram related
     Use embedded viz.js instead of Graphviz. Useful when graphviz isn't installed. Memory is set to 64 MB, if you receive ther error "Cannot enlarge memory arrays" please report this to us.
 [-degree 1 or 2]
     Limit the degree of separation (1 shows less, 2 is default), 1 is a good option for large databases with lots of relationships.
+[-hidelabel]
+    Hide the default label.
--- a/pom.xml
+++ b/pom.xml
@@ -261,7 +261,7 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>8.0.28</version>
+            <version>8.0.33</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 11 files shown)


npetzall:
We have inconsistencies in args and it will be handled in #1031, so will skip anything related that.

npetzall (changes requested):
`schemaspy.hidelabel` should be added to the list of booleans in CombinedDefaultsProvider. It's not obviouse at all, so I'm sorry for that. I'll see if there's a way we can make it more obviouse or if there's a way to automate it.

https://github.com/schemaspy/schemaspy/blob/797a425637b9a5c0fad81bd39c3db4911bad5f4c/src/main/java/org/schemaspy/cli/CombinedDefaultProvider.java#L12

rahulpr27git:
> `schemaspy.hidelabel` should be added to the list of booleans in CombinedDefaultsProvider. It's not obviouse at all, so I'm sorry for that. I'll see if there's a way we can make it more obviouse or if there's a way to automate it.
> 
> https://github.com/schemaspy/schemaspy/blob/797a425637b9a5c0fad81bd39c3db4911bad5f4c/src/main/java/org/schemaspy/cli/CombinedDefaultProvider.java#L12

i have changed the new args format to camel case and added this new field in then `CombinedDefualtProvider.java`

Please review again @npetzall

npetzall (approved):

npetzall:
@rahulpr27git we don't have an issue related to this, so there is no use case/rational related to this change. Could you provide one?

In some sense this change might hurt SchemaSpy recognition/branding/marketing. It goes from a diagram created by SchemaSpy to just a diagram with unknown source.

The marketing is so that anybody will know that this tool exist and is available to the viewer of the diagram, same goes for reports. Hiding this would be a disservice to the viewer.

Anybody viewing the diagram can Google SchemaSpy, find it and use it themselves.

rahulpr27git:
@npetzall I understand.  It's just that some of the clients do not want to see that default label in the ER diagram. HTML templates already have the market label. So I thought it's nice to have this optional args to hide at least on the diagram.

npetzall:
Let's call it a watermark, rather than a label. Since it's servers the purpose of a watermark.

There is no functional need to remove it, it's more of a subjective opinion to remove it. It doesn't cause any harm having it in the diagram, at least not to my knowledge. I believe, and I have no basis for it, that many of the diagrams are re-used in presentations and handmade reports. In those cases the origin would be lost and that could be considered harmful.

When I say harmful, the origin would be lost, the closest parable would be to say that authorship/attribution/copyright should be removed.

You get a different feeling saying that you want to remove a watermark rather than a label. At least I do, it becomes a bit different if you ask the question in a fashion like. "Why do you want to hide the fact that the digram was generated by SchemaSpy?" It might not have been the intention to hide this fact, but it would be the consequence.

Since we don't have an issue regarding this, there has never been a discussion regarding the impact of doing this change.

It would be really good if you could elaborate on the "It's just that some of the clients do not want to see that default label in the ER diagram", since there might be other solutions than removing it, modifying the text, replacing with an icon etc.

But actually removing it, would still be as easy as modifying the generated dot-file and processing it again with Graphviz or even just cropping the image, or using your fork, so we can never stop anyone from "removing" the watermark. I've seen worse watermarks than ours, that actually interferes with that graphics. But we can try to avoid the "wanting to remove it". If we would understand the issue with it better.

rahulpr27git:
Ok. Understood. I will close this PR.

PR #1271, opened 2023-08-18
chore(progress): progress library

npetzall:
* Implemented for Routines pages

### DIFF
--- a/src/main/java/org/schemaspy/progress/ConditionalProgress.java
+++ b/src/main/java/org/schemaspy/progress/ConditionalProgress.java
@@ -1,0 +1,58 @@
+package org.schemaspy.progress;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.InstantSource;
+import java.util.concurrent.atomic.LongAdder;
+
+/**
+ * Proc for reporting progressions conditionally
+ * to avoid always reporting progress.
+ * This is to avoid spamming the log.
+ * Unattended don't want to see every progress
+ * and attended just want to know that the
+ * process isn't hanged.
+ * @param <T> exception thrown by the job
+ */
+public class ConditionalProgress<T extends Throwable> implements Proc<T> {
+  private final ProgressReporter progressReporter;
+  private final FinishCallback finishCallback;
+  private final Condition condition;
+  private final Job<T> job;
+  private final InstantSource instantSource;
+
+  public ConditionalProgress(
+      ProgressReporter progressReporter,
+      FinishCallback finishCallback,
+      Condition condition,
+      Job<T> job,
+      InstantSource instantSource
+  ) {
+    this.progressReporter = progressReporter;
+    this.finishCallback = finishCallback;
+    this.condition = condition;
+    this.job = job;
+    this.instantSource = instantSource;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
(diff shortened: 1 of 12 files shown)


src/main/java/org/schemaspy/progress/ConditionalProgress.java
```diff
@@ -0,0 +1,58 @@
...
+  private final Condition condition;
+  private final Job<T> job;
+  private final InstantSource instantSource;
+
+  public ConditionalProgress(
+      ProgressReporter progressReporter,
+      FinishCallback finishCallback,
+      Condition condition,
+      Job<T> job,
+      InstantSource instantSource
```
jesperolsson-se, line 29: Five parameters is too much, especially since the Javadoc states that all the object does is reducing the reporting.

src/main/java/org/schemaspy/progress/FinishCallback.java
```diff
@@ -0,0 +1,20 @@
+package org.schemaspy.progress;
+
+import java.time.Duration;
+
+/**
+ * Abstraction for reporting or acting
+ * upon the completion of a job, also
+ * it's provided with the number of
+ * progress calls and the duration of
+ * the job
```
jesperolsson-se, line 10: Unclear concern. I bet there's a simpler design, which does not require a callback.

src/main/java/org/schemaspy/progress/Progress.java
```diff
@@ -0,0 +1,12 @@
+package org.schemaspy.progress;
+
+/**
+ * Abstraction to allow a {@link Job} to track progression
```
jesperolsson-se, line 4: Since `Job` knows about `Progress`, `Progress` should not know about `Job`.
The concept is not independent, so perhaps it's better of as a decorator for `Job`?

jesperolsson-se (changes requested):
The design smells.
- The interfaces are tightly coupled to each other and the distinction between them is vague.
- `Progress` and `ProgressReporter` don't seem to be "proper" objects.

Please take some time to explore a design with fewer interfaces and more decrators.

src/main/java/org/schemaspy/progress/ConditionalProgress.java
npetzall: I think so too but it's a later refactor. I've reached my desired output.

src/main/java/org/schemaspy/progress/FinishCallback.java
npetzall: There is but it might require refactors in other places.

I believe that it's more of a task and a job is a series of task, so I hope to eventually reach a state were it's iterators of tasks.

src/main/java/org/schemaspy/progress/Progress.java
npetzall: We are mid refactoring I hope to reach this intermediate state where we can first decouple "stages" later we can think about jobs and tasks.

As I mentioned above. I want the ability to decouple stages and eventually tasks. In short with this intermediary state we can remove the ProgresdListener an maybe think more about independent things that need to be "done" and that they "need" to be tracked.

In the long run we should be able to declare things that should be done and if we choose to track progress or not is a decorator for a task.

jesperolsson-se (approved):

PR #1330, opened 2023-12-05
Update download URL for PostgreSQL JDBC driver

aldahick:
`download.html` 404s, `download/` does not.


npetzall:
Thanks for the update.

PR #1357, opened 2023-12-29
Add note about JAR types

### LINKED ISSUE #1174: "no main manifest attribute" when running from JAR

npetzall:
Fixes #1174


marcelstoer:
Thanks for the effort. However, I don't see the benefit of a new comment-less PR over a squash-merged #1177. I like that GitHub can do this for my projects when I merge PRs. Either way, I just squashed the commits in #1177. Feel free to merge whichever PR you like.

npetzall:
We try to do rebase fast forward to have a cleaner commit history.

Both PR are linked thru the issue. So the amount of comments aren't the issue, they will still exist, and it would have been closed with a reference. But since its been squashed the other will be picked. So this will be closed as it's duplication.

PR #1361, opened 2023-12-31
chore(FileNameGenerator): as decorator

npetzall:
* introducing NameOf as a String adapter
* fix sonarlint issues

PR #1363, opened 2024-01-01
chore: FileNameGenerator as a decorator

npetzall:


npetzall:
Smaller and to the point. Name change as discussed in #1361 with other changes will be in non-review PR.

src/main/java/org/schemaspy/util/naming/NameOf.java
```diff
@@ -0,0 +1,15 @@
+package org.schemaspy.util.naming;
+
+public class NameOf implements Name {
```
jesperolsson-se, line 3: I'd say `NameOf` is a bit too general. `NameOf(Table)` and `NameOf(Database)` makes as much sense as `NameOf(String)`. `StringName` (i.e., a name that needs a string to do its task) would be a narrower name. `Mononym` might also make sense.

jesperolsson-se (approved):

src/main/java/org/schemaspy/util/naming/NameOf.java
npetzall: `NameOf(Database)` and `NameOf(Table)` seems inverted.

Database should have a Name already.

I was inspired by https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/text/TextOf.java

But that is more for Object orientation rather than MicroTyping.

But then `NameFrom(String name)` mig make sense, since it's an adapter or converter.

jesperolsson-se: > I was inspired by https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/text/TextOf.java

The same package has `TextOfDateTime`, `TextOfScalar` and `TextOfString`, which [were extracted](https://github.com/yegor256/cactoos/commit/b8cbd5212d19a57fc6234239788de16d8c5698f1) from it.

> NameOf(Database) and NameOf(Table) seems inverted.
>
> Database should have a Name already.

I agree and think that would be better communicated with the suggested name change.

npetzall: Which of all suggested name changes?

`StringName`, `NameOfString`, `NameFromString`

jesperolsson-se: `NameFromString`

npetzall: > The same package has TextOfDateTime, TextOfScalar and TextOfString, which [were extracted](https://github.com/yegor256/cactoos/commit/b8cbd5212d19a57fc6234239788de16d8c5698f1) from it.

However `TextOf` still contains the `TextOf(String input)` method signature. But it has to do with Text and TextEnvelope which are two different things.

But `NameFromString` and skipping the `NameOf` is fine for our current usage. Since in the end things should have a `Name` and `NameFromString` should be used when interfacing with non-SchemaSpy code.

PR #1374, opened 2024-01-02
chore: add multi-arch docker build

npetzall:


npetzall:
Going on vacation and I don't want to destroy the snapshot tag. So will merge and fix if need when back from vacation

PR #1384, opened 2024-01-14
chore(cli): driverPath as Iterable<Path>

npetzall:
I want help with `Mapped` and `Mapper`

I know that we probably want to avoid `-er` but I want to avoid calling them both `Mapped` since it might cause confusion when reading the code. In some sense `Mapper` is more internal to Mapped. Which means that it shouldn't be over exposed.

```java
new Mapped<>(
  this.driverPath,
  Path::toString
)
```

Since I mainly want to work with Iterable and not the Iterator.

This is more or less the stepping stone that I've talked earlier about. More Iterables and Decorators.

End goal is something like 

```java
new Joined<>(
    new Mapped<>(
        new Filtered<>( //replaces missing path check with inline filtering and logging
            driverPath,
            new WhenFalse<>(
                new PathExist(),
                new LogPath(
                    "Provided -dp(driverPath) is missing: '{}'",
                    LoggerFactory.getLogger(DbDriverLoader.class),
                    Level.WARN
                )
            )
        ),
        SubPaths::new //replaces GetExistingUrls
    )
);
```

If you have Ideas or questions regarding the end goal, we can discuss it.


jesperolsson-se (commented):
I recommend staying with `Mapped`. It brings our attention to the outcome, not the procedure. Worrying about confusion might be a bit pre-mature: worst case we can always make a change later. In the end, there might not be many uses of `Iterator`.

If you want, you can put them in highlight what interface they implement by using different namespaces or a short prefix.

npetzall:
With same name they need to be in different packages and since the primary usage is with in Mapped you would need to use the fully qualified package name for Mapped else there would be a conflict with the class itself Mapped.

But it might just be an issue inside Mapped and not with in any other. My second option would be to sort of follow the 2 char prefix that you've used so IteMapped, ItrMapped.

But as mentioned it will probably just be important to make sure you import the correct one. I believe the primary usage will be of Iterable, since you want to bild graph with it instead of iterator.

PR #1411, opened 2024-02-08
Added LZ4 compression dependency, necessary for Clickhouse compatibility

Danniilpz:
Without this dependency, when trying to run schemaspy with a Clickhouse JDBC against a Clickhouse database, it throws an exception:
`java.sql.SQLException: LZ4 is not supported. Please disable compression(compress=0), modify the algorithm(e.g. compress_algorithm=gzip), or add the missing libraries to the classpath., server ClickHouseNode`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -337,6 +337,11 @@
             <artifactId>nashorn-core</artifactId>
             <version>15.4</version>
         </dependency>
+        <dependency>
+            <groupId>org.lz4</groupId>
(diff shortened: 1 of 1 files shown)


npetzall:
This dependency is not for SchemaSpy but for the driver this can be provided using the `-dp` argument.

Put driver and lz library in a folder and set `-dp` to the folder instead of the driver and library will be available to driver.

Danniilpz:
Thanks, problem solved.

PR #1419, opened 2024-03-03
Add env vars to manage DB connection settings for Docker

apachont:
This PR introduces enhancements to the existing Bash script for the Docker image. The main goal of these changes is to add flexibility to the database connection settings matching the current [command line arguments documentation](https://schemaspy.readthedocs.io/en/latest/configuration/commandline.html#commandline) with ENV variables.

## CLI Args - ENV Vars one-on-one
CLI Arg | ENV Var |
---|---|
--user | USER |
-p | PASSWORD |
-db | DB_NAME |
-host | DB_HOST |
-port | DB_PORT |
-schemas | DB_SCHEMAS |
-t | DB_TYPE|

### Notes
- when DB_SCHEMAS is absent, the script will analyse all schemas in the database.


npetzall:
I wonder if we might have an undocumented and as such untested feature. That I think we can improve. Are you comfortable working in Java?

SchemaSpy has support for reading configurations from the ENV. But I'm pretty sure I added it in haste and it's not well defined.

There would be a small task of just deciding on priority. I think cmd is highest, but if the properties file should be before or after ENV.

And then might want to pre-process the option name before reading to avoid reading random ENVs.

Like looking for "user" but should check S_USER env variable. Then it could be used on any system and not Docker only.

User is a good example since it's often predefined on systems and you don't want to change it. Since other parts of the system might want to use it and in that case it's supposed to be the current running user.

Simples hacked would be change src/main/java/org/schemaspy/cli/EnvDefaultProvider.java

Only respond so `startsWith("schemaspy")`
and make it caps and replace `.` with `_`

Then alla the properties files arguments would work as ENVs but the would be longer like `SCHEMASPY_U` for user since they would have a long prefix.

But that change would be beneficial for the script since it would just be ENV reassignment.

But I wonder if USER will cause problems since it's already defined. Need to look up env variables are case sensitive.

PR #1445, opened 2024-04-25
Update commandline.rst - fix schemaSpec reference

StomproLARL2023:
There was a reference to a "-schemSpec" argument that should be "-schemaSpec".


npetzall:
Thank you for your contribution

PR #1506, opened 2024-07-29
chore(routine): comparison tests

jesperolsson-se:


npetzall:
Have you looked at equalsverifier? We have it as a dependency.

It's been a while since I used it. So not sure if it fits.

jesperolsson-se:
> Have you looked at equalsverifier? We have it as a dependency.
> 
> It's been a while since I used it. So not sure if it fits.

Unfortunately, they don't support `compareTo()`.

npetzall:
To bad since I think it had somewhat of a property based testing approach.

PR #1531, opened 2024-12-28
chore(docker): refactor container image build process

o3bvv:
Hello,

Thank you for the development and maintenance of the project!

This pull-request suggests a bunch of changes to the container image build process, mainly:

1. Switching the image for the drivers-fetching stage from the `curl`-specific image of a fixed version to the latest generic bare `alpine` image, and using built-in `wget` instead of `curl`.
2. Allowing versions of drivers to be specified as build arguments with default values instead of fixed values.
3. Allowing the version of JRE to be overridden during builds.
4. Adding more control to the user creation. User's UID and GID are explicitly set, and can be overwritten during builds, along with the names of the user and its group. This usually becomes handy to deal with permissions for contents of mounted volumes.
5. Installing the `Open Sans` font as a package instead of keeping it as an archive in the repo and copying inside the image. The archive is removed from the repo.
6. Moving the docker entrypoint to the repo root, where the remaining docker-related files reside.
7. Using `tini` as the `init` process inside the container, mainly to handle signals correctly. For example, reacting to SIGTERM while trying to connect to an unreachable DB to avoid waiting until connection's timeout.
8. Squashing `run` commands to reduce the number of resulting image layers.
9. Explicitly marking the output directory as a volume.
10. Cleaning-up the system after installation of packages.
11. Removing the `base` build stage.

> **Trivia**
> I've been working on running [the latest stable Docker image]((https://hub.docker.com/layers/schemaspy/schemaspy/latest/images/sha256-08290ff34c0fa4c1b28af555fbedf755159e42d02651a987e6077cb40a06f1ff)) on my Mac with an M1 chip, but it currently only supports x86-64. To get it running, I had to recompile it locally and took the opportunity to do some refactoring. It seems the next release will include support for arm64 and arm7, so while these changes aren't critical, they could still be quite helpful.

Please let me know if the changes make sense. I'd be glad to follow the contributing guidelines, but the corresponding file is empty, and there are no related pages in the wiki.


npetzall:
Hi, thanks for you contribution. There's a few points to unpack.

1. Switch from a specific to latest, will break the integrity of the build. It will not be reproducible. We have had this issue before. Suddenly the build breaks because there is a change in latest. It's a double edged sword, since you might not get latest security fixes, but you know that it will work because it did so the last time it was built. Suddenly to cut a few extra KB they remove wget or they decide to replace wget with curl.

2. Since we use it with automation, publishing to docker-hub, there is no need for this type of complexity. It also introduces uncertainty, might be a bit nitpicking. But looking at the dockerfile you can't be certain what will be produced since that will be based on arguments provided at build time. What you see might not be what you have. It's a bit of a stretch, you can still inspect the docker image and layers.

3. Same as 2

4. Here we got something, adding UID and GID to the creation would make it more clear what has been created. However the possibility of changing it during build will fall under 2. I think most of the workarounds are based on supplying UID:GID to docker/podman.

5. Is a good fix, the copy of Open Sans is related to when we were using Alpine in the beginning. It wasn't available at the time.

6. I've had discussions with the other maintainer, and I think we came to a conclusion that the root of the repo, should be as clean as possible. But it hasn't been the highest priority. But based on that and since you're a mind reader (no you're not, and it's not an expectation). I would like to move everything container related into a `container` folder. Instead but that is not a big issue.

7. Good

8. Good (driver download image)

9. Good

10. Always good, missing removal of `/var/lib/apt/lists/*`

11. Not too sure about this, but that only has to do with my lack knowledge about caching in bake(will it actually cache each individual layer, and us it as such. Since it's a different build stage at least that would be cached, don't need to re-install everything if only the jar-file changes), and a stupid hope about things running in parallel. Base and driver can run in parallel, since they have no dependencies on each other.

But most of these comments or opinions are based on "users should use the image and not build it themselves". If they are supposed to build it themselves, adding `ARG` would that constitute a public contract for building images and we would need to keep it backwards compatible, or have breaking changes affect the versioning?

I'm gonna think about it and check with the other maintainer. I might start pulling individual commits.

